PR #8, opened 2016-12-02
Allow for nonexistent boundary

ts2909:
When attempting to parse a multipart email where the boundary doesn't exist, enmime gets stuck in an infinite loop.

I have created a test and implemented a fix. Feel free to reject and implement in your own manner! :)


jhillyerd:
Looks great, thank you!

If you send any more PRs, could you send them to the `develop` branch?  I really should add a CONTRIBUTING file to the repo. :)

PR #10, opened 2016-12-06
Add remaining address headers

ts2909:
Added remaining address headers from https://tools.ietf.org/html/rfc2822#section-3.6


jhillyerd:
Excellent, thank you!

PR #12, opened 2016-12-08
Add handling for blank lines in continued header fields

ts2909:
Current handling of continued header fields would terminate if the field had a blank line, ie

```
Subject: Line 1
 Line 2
 
 Line 4
From: test@example.com

This is the body

```
The root part would end up containing:
```
Line 4
From: test@example.com

This is the body

```
Added handling for this and some expanded tests.

### DIFF
--- a/header_test.go
+++ b/header_test.go
@@ -159,13 +159,74 @@ func TestDecodeToUTF8Base64Header(t *testing.T) {
 }
 
 func TestReadHeader(t *testing.T) {
-	prefix := "From: hooman\n"
+	prefix := "From: hooman\n \n being\n"
 	suffix := "Subject: hi\n\nPart body\n"
 
+	data := make([]byte, 16*1024)
+	for i := 0; i < len(data); i++ {
+		data[i] = 'x'
+	}
+	sdata := string(data)
+	_ = sdata
+	/*
+		"Foo: bar\r\n" +
+				"Content-Language: en\r\n" +
+				"SID : 0\r\n" +
+				"Audio Mode : None\r\n" +
+				"Privilege : 127\r\n\r\n"
+	*/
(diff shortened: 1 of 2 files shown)


jhillyerd:
Thanks for the PRs, you are on a roll. :8ball:  My home internet has been offline a couple days, I'll try and dig into these once I get back online.

header_test.go
```diff
@@ -159,14 +159,75 @@ func TestDecodeToUTF8Base64Header(t *testing.T) {
...
-	prefix := "From: hooman\n"
+	prefix := "From: hooman\n \n being\n"
 	suffix := "Subject: hi\n\nPart body\n"
 
+	data := make([]byte, 16*1024)
+	for i := 0; i < len(data); i++ {
+		data[i] = 'x'
+	}
+	sdata := string(data)
+	_ = sdata
```
jhillyerd: I think the _ assignment can be removed, sdata is referenced below

header_test.go
```diff
@@ -159,14 +159,75 @@ func TestDecodeToUTF8Base64Header(t *testing.T) {
...
+	/*
+		"Foo: bar\r\n" +
+				"Content-Language: en\r\n" +
+				"SID : 0\r\n" +
+				"Audio Mode : None\r\n" +
+				"Privilege : 127\r\n\r\n"
+	*/
```
jhillyerd: This looks like it was a temporary comment, can it be removed?

jhillyerd:
Let me know if you'd rather I make the changes myself - I'm not sure how much time you want to put into it.  But I fear if I make the changes and squash the commit, you will not get credit in the commit log.

jhillyerd:
Manually merged w/ fixups.

PR #13, opened 2016-12-08
Missing body

ts2909:
We have a case where we only have the message headers, and not the body. The std library mail.ParseMessage handles this by requiring a separate step to parse the body.

Added handling for this situation along with tests. All existing tests still pass.


jhillyerd:
Squashed & manually merged.

PR #19, opened 2016-12-30
Fix for certain charset duplicate params error

deepakprabhakara:
If a duplicate param is present then parsing of the entire email fails. This fix handles a misplaced ";" in content-type. It also handles empty name in content-type. I have also added a unit test for this scenario.

### DIFF
--- a/envelope.go
+++ b/envelope.go
@@ -2,13 +2,13 @@ package enmime
 
 import (
 	"fmt"
-	"github.com/jaytaylor/html2text"
 	"io"
 	"io/ioutil"
-	"mime"
 	"net/mail"
 	"net/textproto"
 	"strings"
+
+	"github.com/jaytaylor/html2text"
 )
 
 // Envelope is a simplified wrapper for MIME email messages.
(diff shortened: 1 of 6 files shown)


jhillyerd:
It seems like your new error path for missing `;` is not being tested:  https://coveralls.io/builds/9466465/source?filename=part.go#L233 can you add another MIME part to exercise that?

deepakprabhakara:
@jhillyerd Nice catch, didn't know about coveralls. 👍

deepakprabhakara:
@jhillyerd Done. Please review.

deepakprabhakara:
@jhillyerd mime.ParseMediaType is used in several places. I think a better fix is to create a function parseMediaType which then call mime.ParseMediaType and then if that fails call parseBadContentType. This will cover duplicate parameters fixes across Content-Type everywhere (envelope, parts, etc.). Thoughts? I'll update this PR soon.

jhillyerd:
Looks great, thank you!

PR #20, opened 2016-12-30
parse email which send bad charsets like charset="charset=utf-8"

deepakprabhakara:


deepakprabhakara:
I'll submit a test case to cover the else case.

jhillyerd:
Thanks for this, your contributions are very much appreciated!  A couple notes for future pull requests:

- If I'm making changes to `part.go`, I generally try to add tests to its corresponding `part_test.go`, and if possible I write Go-only tests (no .raw file).  While it's good to have some extra cases for envelope_test.go - at some point we will have a thousand .raw files and slow tests. :)  Header parsing is a good use of .raw files though, so no worry with this PR.

- I think enmime is getting fairly close to a beta release.  At that point I will be more strict about PRs needing to go against the `develop` branch instead of `master` - I use https://github.com/nvie/gitflow with default settings.

deepakprabhakara:
Noted, that sounds good to me. You're most welcome. :)

Issue #17, opened 2016-12-22
Help with unsupported charsets


deepakprabhakara:
I have collected a small list of logs for unsupported charsets. Here's the list:-

- ISO646-US
- ISO: Western
- cp936
- cp-850
- 136
- utf-7
- ibm850

As far as I understand ISO646-US is "us-ascii", ISO: Western is "iso8859" and cp936 is "gb2312". Is this correct?

I have been unable to figure out the rest of the encodings although cp-850 and ibm850 seem to be the same.

Does anyone have an idea about these charsets and if they can be supported in this library easily?

deepakprabhakara:
Looks like cp-850 and ibm850 is charmap.CodePage850

jhillyerd:
Looks like you took care of most of these.  Seems like Go does not support UTF-7 out of the box, not sure if that's because Rob Pike invented UTF-8 and doesn't want it.  We could leverage:

https://github.com/cention-sany/utf7

@cention-sany appears to make use of a heavily modified version of the older go.enmime

I'd like to avoid adding a bunch of dependencies if possible, but I'd rather one import than bring in iconv, since it wraps a C library.

deepakprabhakara:
Sounds good to me, I'd avoid iconv as well.

jhillyerd:
Looks like https://github.com/emersion/go-imap/tree/v1/utf7 may be a more idiomatic implementation.  Turns out that UTF-7 was actually *defined* in the [IMAP RFC](https://tools.ietf.org/html/rfc3501#section-5.1.3).

cention-sany:
Just wanna comment that package https://github.com/cention-sany/utf7 is https://tools.ietf.org/html/rfc2152 compliance meanwhile https://github.com/emersion/go-imap/tree/v1/utf7 seem to be https://tools.ietf.org/html/rfc3501#section-5.1.3 which actually two slightly different UTF-7.

jhillyerd:
Good to know, skimming rfc2152, it was specifically designed for internet mail content, so your implementation would be the better choice for enmime.

deepakprabhakara:
@jhillyerd Nice one

jhillyerd:
If I had know it was only going to take 3 lines of code to implement, I might not have waited two years to fix this. 🙄

PR #21, opened 2016-12-30
support more charset mappings

deepakprabhakara:
Fixes issue #17 but unfortunately I don't have any tests for the new charsets added.


jhillyerd:
Looks good.  I don't think there is much value in adding tests for every charset anyways.  Thanks.

PR #23, opened 2017-02-10
Fix and test for miss matched transfer encoding (header says it's quotedprintable when it's not)

chris-garrett:


envelope.go
```diff
@@ -199,6 +199,31 @@ func parseBinaryOnlyBody(root *Part, e *Envelope) error {
...
 }
 
+func parseTextBody(p *Part) ([]byte, error) {
+
+	// use reader that part described
+	allBytes, ioerr := ioutil.ReadAll(p)
+	if ioerr != nil {
+
+		// quoted-printable was not quoted. try raw
+		if strings.Contains(ioerr.Error(), "quotedprintable: invalid unescaped byte") {
```
jhillyerd: In general, it's best to avoid inspecting to the contents of an error string. In this case I don't see an easy way around it though.

https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully (see Never inspect the output of error.Error)

envelope.go
```diff
@@ -199,6 +199,31 @@ func parseBinaryOnlyBody(root *Part, e *Envelope) error {
...
+			reader, err := newCharsetReader("utf-8", p.rawReader)
```
jhillyerd: I may be wrong, but my understanding is that at this point, `p.rawReader` has already been partially or completely consumed.  The plain text portion of the message is lost.  Your unit test below seems to confirm that; I will continue comment there.

envelope_test.go
```diff
@@ -917,3 +917,51 @@ func TestEnvelopeHeaders(t *testing.T) {
...
+	}
+
+	msg := openTestData("mail", "mime-missmatched-encoding.raw")
+	e, err := ReadEnvelope(msg)
+
+	if err != nil {
+		t.Fatal("Failed to parse MIME:", err)
+	}
+
+	if e.Text != "Stuff" {
```
jhillyerd: "Stuff" is not the expected text portion of this message, it should be the ~60 lines of text from the `Content-Type: text/plain; charset="utf-8"` MIME part.

What's happening here is enmime saw that envelope.Text was empty, and down-converted the HTML section to text.  That's not what we want.

envelope_test.go
```diff
@@ -917,3 +917,51 @@ func TestEnvelopeHeaders(t *testing.T) {
 		}
 	}
 }
+
+func TestMissMatchedTransferEncoding(t *testing.T) {
```
jhillyerd: s/MissMatched/Mismatched/

jhillyerd:
I'm going to take a crack at the reader thing, since I've had to implement a couple io.Readers for enmime already...

jhillyerd:
Ok, I've added a quoted-printable cleaning reader in 7706cdf, and isolated the text portion of your slack test case in 7e461ff.  Feel free to open a bug if you find more q-p problems.

envelope.go
chris-garrett: Noted ty.

envelope.go
chris-garrett: Yes you are right. I totally know better.

chris-garrett:
Thank you for entertaining my rushed, poorly thought out PR. I will test your changes and let you know if anything crops up. Cheers!

PR #24, opened 2017-02-10
Fix and test for text only bodies that are plain text

chris-garrett:


jhillyerd:
Thank you!  I realized after looking at the code that you were basically working around a bug in the isBinaryBody method, so I fixed the bug and made a couple tweaks to your test in 8c6ab1211956d40dc67a31064863ee7ce0128e33. You also helped me discover #25 :)

PR #26, opened 2017-02-17
Handle unencoded equals in quoted printable sanitizer

chris-garrett:
I ran into an issue today with the sanitizer. Another email from slack came through but this time it contained a url with query parameters. The '=' passes through to qp reader and fails. Here is my first pass. Thoughts?

### DIFF
--- a/part.go
+++ b/part.go
@@ -86,7 +86,7 @@ func (p *Part) buildContentReaders(r io.Reader) error {
 	encoding := p.Header.Get(hnContentEncoding)
 	switch strings.ToLower(encoding) {
 	case "quoted-printable":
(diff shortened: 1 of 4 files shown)


jhillyerd:
My first thought is "man Slack kinda sucks!" ;)  I like your approach here though, nice!  Could you add some test to exercise the three red error paths in:

https://coveralls.io/builds/10212874/source?filename=quotedprint.go

Should be as easy as adding a few more cases to this test table:

https://github.com/jhillyerd/enmime/blob/master/quotedprint_test.go#L21

chris-garrett:
Yes, definitely.

jhillyerd:
Thank you!

PR #29, opened 2017-06-19
Support for binary body without 'Content-Disposition' header.

davrux:
The following mail was interpreted as non binary message body:

To: bob@test.com
From: alice@test.com
Subject: Test
Message-ID: <56A0AA5F.4020203@test.com>
Date: Thu, 21 Jan 2016 10:52:31 +0100
MIME-Version: 1.0
Content-Type: application/pdf;
 name="doc.pdf"

iVBORw0KGgoAA...

This is an attachment only mail without a Content-Disposition-Header.
Added Support to detect this.


jhillyerd:
Thank you!

Issue #42, opened 2017-11-22
Email's attachment won't be exported correctly.


leibiao:
I try to export an email's attachment(Zip archive), but that exported file could not be unzip successfully.I compare this file to which exported by emali client, like follows:
`50 4B 03 04 14 00 08 00 08 00 EF BF BD 02 29 4A` -->export by enmime
`50 4B 03 04 14 00 08 00 08 00 C2 02 29 4A 00 00` -->export by email client
It's clearly that the byte 'C2' transfers to 'EF BF BD 02', I also check the implementation of Part's Read method:
` 
return p.utf8Reader.Read(b)  
`
OK, maybe it causes the issue when I need to read a binary file. So can export a METHOD for read the decoded binary content?

jhillyerd:
I was able to write a test to reproduce this in 3dae741f77ff413f6e0dc7cd18537ca4e7563f67 - the only catch is that the attachment had to have a charset encoding header - which a ZIP file should not.  Can you check the email that triggered the bug and confirm?

leibiao:
Yes. The reason you point out triggered the bug.Because the zip file's name is not english(other language), this header is used for the name instead of the content.

Issue #34, opened 2017-10-31
Base64 decoding error


Requaos:
I have an email attachment that is being mishandled. Its a curious thing, probably some funky counting issue with the padding chars.

After some investigation I found that during the buffer sizing, or resizing, it is dropping the last quad of the base64 string... string input length is 12992, but only 12988 characters are being fed to the decoder OR the receiving buffer is sized to 9615 instead of the 9617, causing the last two bytes to be trimmed from the tail of the byte stream.

I don't know precisely which copy operation is causing the miscount and to be fare I have a service running with your processing engine and this fringe case is one out of 60,000 emails with an average of 3 attachments per. 

other files process fine... perhaps we could take this offline to exchange my sample file because it contains confidential information and review it in attempt to reproduce the issue with a benign file exhibiting the same features.

fyi, I am able to decode the blob accurately using this:
```
f, _ := ioutil.ReadFile("base64.txt")
base64Bytes, _ := base64.StdEncoding.DecodeString(string(f))
file, _ := os.Create("decoded.xlsx")
defer file.Close()
writer := bufio.NewWriter(file)
writer.Write(base64Bytes)
defer writer.Flush()
```

Thoughts?

jhillyerd:
Apologies for not responding, overloaded at work.  This may be fixed in #43 (not yet merged), the tl;dr is enmime is incorrectly attempting to UTF-8 decode binary content.

Requaos:
Cool, I have a local repo setup to reproduce this bug, so when the fix is pushed to an issue branch, I can test to verify that it is resolved before merging with master

jhillyerd:
I've merged #43 into `develop` branch; give it a shot and let me know.

Requaos:
Yeah, it's at work so I'll let you know on Monday afternoon

jhillyerd:
Everything is now merged to master

jhillyerd:
Going to close this, please re-open if you are able to repro the issue.  Alternately, the email listed on my profile is private, so safe to send attachment to.  You may want to ping this bug after emailing in case it goes into spam.

PR #43, opened 2017-11-22
Don't use CharsetReader when Part is attachement

leibiao:
For this [issue](https://github.com/jhillyerd/enmime/issues/42)

### DIFF
--- a/part.go
+++ b/part.go
@@ -103,7 +103,7 @@ func (p *Part) buildContentReaders(r io.Reader) error {
 	}
 	p.decodedReader = contentReader
 
-	if valid {
+	if valid && !isAttachment(p.Header){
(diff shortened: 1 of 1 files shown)


jhillyerd:
This is a *really* nice find.  Thank you! I was really dreading digging into bug #34 but I have a feeling it has the same root cause as #42 you filed.

I've been very busy at work and it looks like I've let a back log of PRs pile up, will try and spend some time this weekend getting them merged.

Would it be possible for you to add a corresponding test to part_test.go? You could create a new testdata/parts file based on https://github.com/leibiao/enmime/blob/master/testdata/parts/multibase64.raw .  Please try to limit the size of the test base64, I don't want to bloat the repository unnecessarily.

Issue #46, opened 2017-12-22
Should we make enmime more  fault-tolerant？


leibiao:
Recently I made a little test using following content with **enmime**
```
Content-Type: multipart/mixed; boundary="Enmime-Test-100"

--Enmime-Test-100
Content-Transfer-Encoding: base64
Content-Type: text/html; charset=utf-8

PGgxPlRlc3QhPC9oMT4K
--Enmime-Test-100
Content-Transfer-Encoding: base64
Content-Type: application/octet-stream; name="test.txt.gz"; charset="us-ascii"
Content-Disposition: attachment; filename=test.txt.gz

H4sICG2/PFoAA3Rlc3QudHh0ACtJLS7hAgDGNbk7BQAAAA==

--Enmime-Test-100--
```
As you can see, when I append a '=' to the html part's content `PGgxPlRlc3QhPC9oMT4K`, it's a malformed base64 encoded string and will abort the work of **enmime**.Ok, I saved this modified content and opened it with macOS's Mail app, I also can get the correct attachment though the html part is displayed incorrecly.
Another example is, when I insert a speciall byte into fourth from the bottom of attachment's content  like follows.
```
41 3D 3D    A==     ------->     11 41 3D 3D     .A==
```
Obviously, it is also a malformed base64 encoded string, but it can be extracted with **enmime** .The key problem is the extracted file can't be uncompressed while extracted with another tools like  Mail.app,Thunderbird and python's built-in package can be uncompressed correctly.

jhillyerd:
Yes, my goal for enmime is to make it parse most reasonable emails, and aggregate the parsing errors so that tools like [inbucket](http://www.inbucket.org/) can display them; you can see this work started in #1 

We even have a cleaner in base64.go that attempts to work around some bugs I found with Go's own base64 decoder when first building enmine.  It was never wired into the error aggregation system.

It would make sense to add an `Errors` slice to the `base64Cleaner` struct, and an `errorMalformedBase64` const to `errors.go`, plus whatever other plumbing is required to surface those errors to the consumer.

PR #47, opened 2017-12-23
Improve the tolerance of base64Cleaner

leibiao:
Inspired by python, for this [issure](https://github.com/jhillyerd/enmime/issues/46)

### DIFF
--- a/base64.go
+++ b/base64.go
@@ -41,7 +41,7 @@ func (qp *base64Cleaner) Read(p []byte) (n int, err error) {
 	buf := qp.buf[:size]
 	bn, err := qp.in.Read(buf)
 	for i := 0; i < bn; i++ {
-		switch base64CleanTable[buf[i]&0x7f] {
+		switch base64CleanerTable[buf[i]&0x7f] {
 		case -1:
 			// Strip these
 		default:
--- a/base64.go
+++ b/base64.go
@@ -4,6 +4,18 @@ import (
 	"io"
 )
 
+// something inspired by table_a2b_base64 in binascii module of python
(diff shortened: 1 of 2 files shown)


jhillyerd:
Thanks, I hadn't seen that python table before.  I like this idea, but I don't like that we are not capturing any errors to send back to the software using enmime.

From my comment on #46 - It would make sense to add an Errors slice to the base64Cleaner struct, and an errorMalformedBase64 const to errors.go, plus whatever other plumbing is required to surface those errors to the consumer.

Perhaps we can keep the original `case ' ', '\t', '\r', '\n':` statement, and add '=' to it, allowing all those to be stripped silently.  But any other characters returning -1 would trigger an error to be added to the list.

Let me know if you are interested in pursuing this, otherwise I can merge this PR into a feature branch and add the error part myself.

leibiao:
Maybe we can add a switch that controls the strategy of **enmime**.For example, if use _**Strict**_ mode, **enmime** will generate more errors when meets malformed strings.And use **_Tolerance_** mode, **enmime** will try it best to rebuild the content while only generates some necessary errors.   
BTW, it's common to analyze malformed spam in my work.I'd like to meet this switch added to **enmime**.  :)

jhillyerd:
Take a closer look at https://github.com/jhillyerd/enmime/blob/master/error.go#L19

We don't need the flag, as the client can filter on `Severe` or `Name` to only examine errors it cares about.  Or it can completely ignore the Part/Envelope.Errors slice.  In which case it only need to pay attention to the hard Go `error` returned from `ReadEnvelope` to know that parsing completely failed.

PR #54, opened 2018-01-27
encode: FileName must be quoted

tgulacsi:
mime.FormatMediaType returns "" if the value is not a quoted-string as
per RFC 2616.

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -79,7 +79,7 @@ func (p *Part) Encode(writer io.Writer) error {
 		}
 		mt := mime.FormatMediaType(p.ContentType, param)
 		if mt == "" {
-			// fmt.Printf("cannot encode %q with %q", p.ContentType, param)
+			// there was some error, FormatMediaType couldn't encode it with the params
 			mt = p.ContentType
 		}
 		p.Header.Set(hnContentType, mt)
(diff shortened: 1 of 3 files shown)


jhillyerd:
Thanks for this, I'm surprised anybody has even seen this branch yet!

Can you add a golden file test to prove the quoting is working?  I'd consider cloning https://github.com/jhillyerd/enmime/blob/feature/builder/encode_test.go#L33 then change the filename in the test.DiffGolden call.

You'll need to `touch` the file, then run `go test -update` to generate it, inspect the output to make sure it's good then commit.

tgulacsi:
Done, thanks!

jhillyerd:
Thanks! Your code is definitely performing better than my original code, as running that test against it yields empty headers:
```
Content-Disposition:
Content-Id: <mycontentid>
Content-Type:

WklQWklQWklQ
```

But I noticed that the filename looks double-quoted now: `filename="\"\\u00e1rv`; shouldn't that be: `filename="\u00e1rv` or am I misunderstanding?

tgulacsi:
You're right. But who does the second quoting? I'll try to find it.

James Hillyerd <notifications@github.com> ezt írta (időpont: 2018. jan.
27., Szo, 22:46):

tgulacsi:
https://www.w3.org/Protocols/rfc2616/rfc2616-sec2.html#sec2.2 
says some chars must be escaped with \, https://golang.org/src/mime/mediatype.go#L61 says \ will be doubled, but chars > 127 (non-ASCII) will induce "". So I cannot see any way to avoid double-\...

jhillyerd:
I think this is a better than a blank filename for now.  I have some ideas of how we might improve it, but I also have a bunch of code I want to merge in; so I will accept this and file a bug to discuss possible improvements.

PR #65, opened 2018-03-28
Missing Boundary on all Mime Parts

pedrinimm:
Currently Parts insider envelope do not contain their boundary.

This is an issue when you have a multipart/alternative email inside a multipart/mixed. Because the if you want to retrieve the boundary of the html part inside the multipart/alternative you cant because the "parent" of this parts is its sibling txt. This one has an empty boundary. 

IMHO every mime part should be aware of its own boundary. 

This is added on part.go line 336

In addition I added some duplication functions just because they are useful when working with multiple copies of the same envelope.

### DIFF
--- a/part.go
+++ b/part.go
@@ -229,6 +229,39 @@ func (p *Part) buildContentReaders(r io.Reader) error {
 	return err
 }
 
+// DuplicatePart returns a duplicate of the current Part
+func (p *Part) DuplicatePart() *Part {
+	if p == nil {
+		return nil
+	}
+	newPart := &Part{
+		p.PartID,
+		p.Header,
+		nil,
+		p.FirstChild.DuplicatePart(),
+		p.NextSibling.DuplicatePart(),
+		p.Boundary,
+		p.ContentID,
+		p.ContentType,
+		p.Disposition,
+		p.FileName,
+		p.Charset,
+		p.Errors,
+		p.Content,
+		p.Epilogue,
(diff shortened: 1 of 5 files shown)


part.go
```diff
@@ -296,9 +329,11 @@ func parseParts(parent *Part, reader *bufio.Reader) error {
 		parent.AddChild(p)
 		if p.Boundary == "" {
 			// Content is text or data; build content reader pipeline.
-			if err := p.buildContentReaders(bbr); err != nil {
+			err := p.buildContentReaders(bbr)
+			if err != nil {
 				return err
 			}
+			p.Boundary = parent.Boundary
```
jhillyerd: I don't agree with this change; Boundary is a header field belonging to the parent object that it is defined on, it doesn't make sense to duplicate it on the children.  Especially not using the same Part field.

> if you want to retrieve the boundary of the html part inside the multipart/alternative you cant because the "parent" of this parts is its sibling txt.

As far as I know, your statement above is incorrect: siblings are not parents of each other. The Part.Parent should be pointing to the same parent (multipart/*) for all the siblings.   If that's not the case, then it sounds like a bug.

Let's remove this change and keep this PR scoped to the Clone and Test helper changes.  Then file a bug for the parent/boundary thing, possibly with some code or a unit test to illustrate the problem or use case.

jhillyerd:
Closing in favor of #66, thanks for the PR.

PR #66, opened 2018-04-03
allow to clone envelopes

nerdlich:
This kind of supersedes PR #65 (I don't have access to pedrinimm's branch). Will look into the mentioned boundary issue separately.


jhillyerd:
Thanks guys.

PR #68, opened 2018-04-14
Fix parseBinaryOnlyBody by adding Content

ostcar:
Currently, the `Content` of a BinaryOnlyMail is empty. It is possible to access the content with `env.Inlines[0].Content`, but this does not work in any case.

In my use case, I save all images from incoming messages. I do this with `DepthMatchAll()`. It returns a list of `Part`-objects. But in a BinaryOnlyMail there is no link between the `Part` and `env.Inlines[0]`. So without this patch, I see no way to access the content.

### DIFF
--- a/envelope.go
+++ b/envelope.go
@@ -207,8 +207,9 @@ func parseBinaryOnlyBody(root *Part, e *Envelope) error {
 	// Add our part to the appropriate section of the Envelope
 	e.Root = NewPart(nil, mediatype)
 
-	// Add header from binary only part
+	// Add header and content from binary only part
 	e.Root.Header = root.Header
+	e.Root.Content = root.Content
 
 	if root.Disposition == cdInline {
 		e.Inlines = append(e.Inlines, root)
--- a/envelope_test.go
+++ b/envelope_test.go
@@ -654,6 +654,10 @@ func TestAttachmentOnly(t *testing.T) {
 		if len(e.Root.Header) < 1 {
 			t.Errorf("No root header defined, but must be set from binary only part.")
 		}
+		// Check, that the root part has content
+		if len(e.Root.Content) == 0 {
+			t.Errorf("Root part of envelope has no content.")
+		}
 	}
 }
 


jhillyerd:
I need to think about this a little.  In cases where Inlines is empty, will be in Attachments instead - the disposition headers control this; you can see it being tested in the unit test you modified.

If we start adding the Content to the root part, it may break existing users, but it may be more accurate if that root part also has the content-type of the attachment.

ostcar:
Can you give me a tip, what I can do to solve this problem in my code without this fix?

Currently [this line](https://github.com/ostcar/mailimage/blob/43a52dbd7db25ee3761551b65ce56e98f234e943/insert.go#L60) returns the `part` without the content. The only way I see to get the content would be, if the argument of the function would be an `Envelope` instead of the `Part` and that I write:
```go
imageParts := envelope.Root.DepthMatchAll(imageMatcher)
...
if len(imageParts[0].Content) == 0 {
  content = envelope.Inline[0].Content
} else {
  content = part.Content
}
```
But this seems like a hack.

This problem is very likely in my usecase, because apple-mail sends BinaryOnlyMail, if you only add an Image and no text. At least one of my users uses apple-mail

jhillyerd:
What I was thinking was something like:
```go
if len(imageParts[0].Content) == 0 {
  if len(envelope.Inlines) > 0 {
    content = envelope.Inlines[0].Content
  } else if len(envelope.Attachments) > 0 {
    content = envelope.Attachments[0].Content
  }
} else {
  content = part.Content
}
```

but I agree, it's hacky.  Can you share the code for `imageMatcher`?  That may help me visualize things better.

ostcar:
The code for the `imageMatcher` is [here](https://github.com/ostcar/mailimage/blob/43a52dbd7db25ee3761551b65ce56e98f234e943/insert.go#L46)

```go
func isAllowed(contentType string) bool {
	for _, format := range AllowedFormats {
		if ("image/" + format) == contentType {
			return true
		}
	}
	return false
}

func imageMatcher(part *enmime.Part) bool {
	return isAllowed(part.ContentType)
}
```

jhillyerd:
Spent some time staring at the code, and now I agree that it makes sense to add content.

I also realized that most of what parseBinaryOnlyBody does is redundant.   Removing it will likely also fix this issue, but its still worth merging this PR for the unit test change.

Thanks!

Issue #11, opened 2016-12-07
Access all headers from Envelope


ts2909:
With the later changes, the headers field of the Envelope has been made private. I have need of all headers in the email and there does not seem to be a way of doing this without knowing ahead of time what they are, and using the GetHeader method.

Do you have a recommended way of doing this, ie perhaps a new HeaderKeys method, make headers public, or a new GetRawHeader method, or perhaps there is something I'm missing?

jhillyerd:
The headers can be accessed through the Root Part: https://github.com/jhillyerd/enmime/blob/master/part.go#L18

I may end up making them public in the Envelope again to support encoding new mail, not sure yet...

jhillyerd:
Thinking along these lines, once thing I noticed when messing around with the example test for the [cmd utils](https://github.com/jhillyerd/enmime/blob/master/cmd/utils.go#L63) package is that textproto.MIMEHeader is a map, so it loses ordering once read.

The way I designed inbucket, it shows the full message source, which it obviously has access to.  Other enmime clients may want to just display the header block, and it seems silly to make them parse that section out of the raw email when enmime already does that.

Do you think it would be useful to have a way to get the raw message header as a block of text (or slice of lines)?

jhillyerd:
Also:  Given you've been in the enmime source, and it wasn't obvious that Envelope.Root held the headers, do I need to better document the Envelope/Part relationship?

I saw a quote the other day that developers are mostly looking for example code to cut & paste when trying out libraries, so maybe I should put together some more examples and link to them from the README.  I'm not sure many developers even know the [current example](https://godoc.org/github.com/jhillyerd/enmime#example-package) even exists.

ts2909:
Funny you mention just needing the headers, I just created a pull request with a fix to parse only the headers. We sometimes pull only the headers from IMAP and just need to parse and decode this.

In our case we don't need to maintain the order of the headers, but I suppose there could be a use case for this, but it's difficult when the standard library functions don't allow this!

One of the first things I do with golang code is look at the example, so I was aware of this.. There isn't anything in there about the Root, or the message headers, besides the GetHeader function, which when you look at it, accesses the private header field. It's not quite clear the Root headers and the same as the Envelope field.

But yeah, good work, it's exactly what I'm looking for, particularly the new work on the Warnings. We need to be as liberal as possible when decoding mail, it's not really an option to just fall over, we need to extract as much as possible.

[...]

PR #69, opened 2018-05-23
Add Envelope.GetHeaders()

allenluce:
This provides (roundabout) access to the `textproto.MIMEHeader` field of the `Envelope` struct. It's handy for various situations, like telling if there are optional DKIM or SPF headers.

### DIFF
--- a/envelope.go
+++ b/envelope.go
@@ -23,6 +23,18 @@ type Envelope struct {
 	header      *textproto.MIMEHeader // Header from original message
 }
 
+// GetHeaderKeys returns a list of header keys seen in this message. Get
+// individual headers with `GetHeader(name)`
+func (e *Envelope) GetHeaderKeys() (headers []string) {
+	if e.header == nil {
+		return
+	}
+	for key := range *e.header {
+		headers = append(headers, key)
+	}
+	return headers
+}
+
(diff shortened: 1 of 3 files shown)


jhillyerd:
Thanks, was just talking about this in #11.  Looks like the linter does not like:
```
/enmime/envelope.go:34:11:
should omit 2nd value from range; this loop is equivalent to `for key := range ...
Found 1 lint suggestions; failing.
```

Will try and take a closer look at this PR by the end of the weekend.

allenluce:
That should be taken care of now. Do let me know if you'd prefer a different method name (per discussion on #11).

jhillyerd:
I know `Get`ter naming is frowned upon in Go community, but since we already have `GetHeader`, I think `GetHeaderKeys` would be the best method name.

I'm on the fence as to whether it should sort its results or not... probably best to leave it the way it is and see if anyone complains about order randomization.

Everything else looks good to me.

allenluce:
Does this need anything else?

jhillyerd:
Looks like no, I got confused by the title and was waiting for you to change the method to GetHeaderKeys, but I see that is already the name used in the code.  Thanks!

PR #71, opened 2018-06-18
Bugfix HTML charset.

davrux:
A chaset maybe set in header of HTML only emails, but not set in HTML
part. If a charset is set on HTML only email, use it to convert
content.

If Content-Type: text/html;charset="iso-8859-1" is set in mail header, but HTML body does not contain meta Content-Type, the body is not converted to utf-8.


davrux:
Added test.

jhillyerd:
Thanks, looks good!

PR #73, opened 2018-06-29
Fix for multipart/mixed with inline disposition.

davrux:
If mail header is

   Content-Type: multipart/mixed;...
   Content-Disposition: inline

the resulting Envelope has an additional inline attachment with zero
bytes and a content type multipart/mixed. Changed test, so that
inlines with "multipart/" as content type prefix are not added to
inlines slice.

### DIFF
--- a/envelope_test.go
+++ b/envelope_test.go
@@ -492,6 +492,40 @@ func TestParseHTMLOnlyInline(t *testing.T) {
 	}
 }
 
+func TestParseInlineMultipart(t *testing.T) {
+	msg := test.OpenTestData("mail", "inlinemultipart.raw")
+	e, err := enmime.ReadEnvelope(msg)
+	if err != nil {
+		t.Fatal("Failed to parse MIME:", err)
+	}
+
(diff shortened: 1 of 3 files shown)


envelope_test.go
```diff
@@ -492,6 +492,40 @@ func TestParseHTMLOnlyInline(t *testing.T) {
...
+		t.Errorf("len(e.Errors) got: %v, want: 0", len(e.Errors))
+	}
+
+	want := "Simple text."
+	if !strings.Contains(e.Text, want) {
+		t.Errorf("Downconverted Text: %q should contain: %q", e.Text, want)
+	}
+
+	if len(e.Inlines) != 1 {
+		t.Error("Should have no inline, got:", len(e.Inlines))
```
jhillyerd: It looks like this test is expecting one inline, but error text indicates otherwise.

davrux:
Fixed ;)

jhillyerd:
Thanks!  I'm planning on doing a 1.0 release soon, so this ought to hit master in a month or so.

PR #74, opened 2018-07-07
Use filepath and not url when dealing with files

### LINKED ISSUE #57: cmds should use filepath not path pkg
`path` is for URLs, `filepath` is for OS file path manipulation

```
cmd/mime-dump/mime-dump.go:7:   "path"
cmd/mime-dump/mime-dump.go:26:  basename := path.Base(os.Args[1])
cmd/mime-extractor/mime-extractor.go:9: "path"
cmd/mime-extractor/mime-extractor.go:44:        basename := path.Base(*mimefile)
cmd/mime-extractor/mime-extractor.go:59:                newFileName := path.Join(*outdir, a.FileName)
```

stuartskelton:
This fixes  #57.

was fairly straight forward, I have tested the changes compile and the code builds and runs, on Mac.

### DIFF
--- a/cmd/mime-dump/mime-dump.go
+++ b/cmd/mime-dump/mime-dump.go
@@ -4,7 +4,7 @@ package main
 import (
 	"fmt"
 	"os"
-	"path"
+	"path/filepath"
 
 	"github.com/jhillyerd/enmime"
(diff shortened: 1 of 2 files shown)


jhillyerd:
Thanks for this!

PR #84, opened 2018-11-12
use iso-8859-1 charmap over windows-1252 when implicitly defined

requaos:
Had some issues with this charset being decoded as Windows-1252:
[Information Here](https://www.i18nqa.com/debug/table-iso8859-1-vs-windows-1252.html)


requaos:
Turns out this will not solve my problem, since the batch of problem cases are multipart and have `inline` in the content-disposition header, which triggers the `detectAttachmentHeader` function to return `true`, so no `coding.NewCharsetReader` is ever invoked... I'll put this PR on hold until I find a better solution.

requaos:
```--00B0FEEE_message_boundary
Content-type: text/plain; charset=iso-8859-1
Content-Transfer-Encoding: Quoted-printable
Content-Disposition: inline
Content-Description: Message text```

PR #85, opened 2018-11-13
Some mail clients default the `Content-Disposition` to `inline` for plain text section

requaos:
This PR attempts to make the disposition of attachment versus text section more reliable for the proper charset decoding of `iso-8859-1` text sections


jhillyerd:
Can you make sure you are basing your work off the development branch; that's where PRs should go.

Looks like it broke one of the unit tests.

requaos:
I will change the target branch, and I found the broken test, it was trying to decode the html text portion twice.

requaos:
rebased against `develop` branch and changed merge target, tests are passing and code coverage has increased due to commenting out redundant charset conversion on html text part

requaos:
What is your preference on commented code? leave in-place with comment for why it has been commented out? or just remove it and have meaningful commit message?

jhillyerd:
Thanks.

Re commened-out code: if we don't anticipate needing it in the very near future, I think removing it is best.

Will take me some time to review (likely this weekend), want to make sure I understand the change before merging.

requaos:
I found that text bodies commonly have `inline` listed as the content disposition, but without any parameters if it is a visible text body... its not perfect because mail clients do the strangest things at times, but this change allows the visible text portions to go through charset decoding earlier (some where even being skipped, in the case of `text/plain` visible portions with `Content-Disposition: inline`

Issue #81, opened 2018-10-24
Declared charset cannot always be trusted


nerdlich:
Unfortunately, you cannot always trust the charset that is defined in a Content-Type header. We've seen cases where enmime essentially corrupts part.Content by decoding a declared gbk charset where the data in fact was utf-8. 

Instead of taking for granted whatever is declared in the header, one could ask a charset detection library like https://github.com/saintfish/chardet. I'll see if I can prepare a PR. Let me know what you think.

jhillyerd:
Sounds like a good idea to me.  I do wonder, how would we decide which to trust?

nerdlich:
The results of this library have a "confidence" field. No idea how that's calculated though ;)

To be really sure we could do a full decode - encode cycle with the detected charset and compare the input and output bytes. But that's maybe too much overhead.

jhillyerd:
Confidence scoring sounds like it should work for our use case.

I agree that doing multple decode/encode cycles would be too expensive to do all the time in enmime.  What may make more sense is to build an external tool that would do that test over a large corpus of email (if you have one, I need to find one...), allowing us to input a confidence score threshold, then we could see what the failure rate was given that threshold.

nerdlich:
I've tried the Enron data set (https://www.cs.cmu.edu/~enron/), but not only is the data very old, it also doesn't contain any interesting charsets at all, mainly us-ascii, utf-8, iso-8859-1. 

I might be able to gather a body of emails but only for a private test, not to share publicly.

nerdlich:
Did some more tests now with a body of ~3GB of emails received in China, Japan, Korea, Pakistan, UAE, Russia, Bulgaria, Greece, hence lots of potentially problematic charsets involved. You get a lot of false positives and boring stuff (e.g. UTF-8 instead of ISO-8859-1) with lower confidence scores, but with confidence >90, results look very promising. The majority of cases is GBK or GB2312 detected as GB18030 which according to my understanding is a superset of the two former, so in most cases probably not a real mismatch, but also not wrong to use GB18030. But there are also a few true positives like GB2312 correctly detected as UTF-8 or us-ascii correctly detected as ISO-2022-JP.

To leave all options open for the users of enmime we could also keep both, "OrigContent" and "OrigCharset" (as specified), and "Content" and "Charset" (as detected) in the part struct.

jhillyerd:
[...]

PR #87, opened 2018-11-14
detect charset of email parts

nerdlich:
This implements issue #81. Let me know if you think the content should also be decoded with the original charset, such that users can choose which one they want to use.

### DIFF
--- a/part_test.go
+++ b/part_test.go
@@ -328,7 +328,7 @@ func TestMultiMixedParts(t *testing.T) {
 		Parent:      test.PartExists,
 		NextSibling: test.PartExists,
 		ContentType: "text/plain",
-		Charset:     "us-ascii",
+		Charset:     "ISO-8859-1",
 		PartID:      "1",
 	}
 	test.ComparePart(t, p, wantp)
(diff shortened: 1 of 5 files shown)


part_test.go
```diff
@@ -328,7 +328,7 @@ func TestMultiMixedParts(t *testing.T) {
 		Parent:      test.PartExists,
 		NextSibling: test.PartExists,
 		ContentType: "text/plain",
-		Charset:     "us-ascii",
+		Charset:     "ISO-8859-1",
```
dcormier: I'm not so sure about this. Looking at the part in question (`multimixed.raw`), it appears that it is indeed `us-ascii`. Sure, `ISO-8859-1` (should that be lowercase?) will work, but so would `utf-8`.

Thoughts, @jhillyerd?

jhillyerd: I think they are the same when there is no 8-bit characters present.  Based on [this table](http://www.columbia.edu/kermit/ascii.html) it seems the only 8-bit characters in us-ascii are control codes that wouldn't be present in an email.

I think that means there would be zero impact to how a `us-ascii` detected as `iso-8859-1` message is decoded to UTF-8, which is our goal.

If we do start seeing problems, we can look at work-arounds or forking the [chardet package](https://github.com/gogs/chardet) (it's ancient).

PR #88, opened 2018-11-15
ensure that media parameters with special characters are quoted

requaos:


requaos:
WIP: still sorting out the precedence of duplicate parameters, missing semicolon and lack of quotes around strings with special characters therein

requaos:
Ready for review

jhillyerd:
Nice, thanks.

PR #89, opened 2018-11-16
Resolves an issue parsing a malformed Content-Type header

dcormier:
An email was received with this Content-Type header:

```
Content-Type: ;
	name=""
```

This change makes it so this MIME part can be parsed successfully.

### DIFF
--- a/header.go
+++ b/header.go
@@ -27,6 +27,9 @@ const (
 	ctTextPlain        = "text/plain"
 	ctTextHTML         = "text/html"
 
+	// Used as a placeholder in case of malformed Content-Type headers
+	ctPlaceholder = "x-not-a-mime-type/x-not-a-mime-type"
+
 	// Standard Transfer encodings
 	cte7Bit            = "7bit"
 	cte8Bit            = "8bit"
@@ -210,6 +213,9 @@ func parseMediaType(ctype string) (mtype string, params map[string]string, err e
 			}
 		}
 	}
+	if mtype == ctPlaceholder {
+		mtype = ""
+	}
 	return mtype, params, err
 }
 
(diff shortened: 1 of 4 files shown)


testdata/parts/badattach.raw
```diff
@@ -0,0 +1,6 @@
+Content-Type: ;
+	name=""
+Content-Transfer-Encoding: base64
+Content-Disposition: attachment;
+	filename=""
+
```
dcormier: This is actually the entirety of the MIME part that was received in the email the resulted in this PR.

header.go
```diff
@@ -216,12 +222,20 @@ func fixMangledMediaType(mtype, sep string) string {
...
-				// Ignore repeated parameters.
-				continue
+	for i, p := range parts {
+		switch i {
+		case 0:
+			if p == "" {
+				// The content type is completely missing. Put in a placeholder.
+				p = ctPlaceholder
+			}
+		default:
```
dcormier: The real change is above this comment. The change below is just an indenting change to put it in the `default` case of the added `switch`.

jhillyerd:
Thanks for this.  Will try to review this weekend, but there are a few in the hopper before you.

It may make sense to return ErrorMissingContentType and have it added to Part.Errors so that we can surface the warning to interested clients.

dcormier:
> Thanks for this. Will try to review this weekend, but there are a few in the hopper before you.

Understood. Thanks for getting back to me.

> It may make sense to return ErrorMissingContentType and have it added to Part.Errors so that we can surface the warning to interested clients.

OK. I'll take a look at that.

dcormier:
> It may make sense to return ErrorMissingContentType and have it added to Part.Errors so that we can surface the warning to interested clients.

OK, now the part will have that added as a warning.

testdata/parts/badattach.raw
dcormier: Done.

part_test.go
```diff
@@ -740,3 +740,47 @@ func TestClonePart(t *testing.T) {
 	clone := p.Clone(nil)
 	test.ComparePart(t, clone, p)
 }
+
+func TestBadAttach(t *testing.T) {
```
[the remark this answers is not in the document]
dcormier: Done.

dcormier:
Alright, I believe I have addressed your comments.

jhillyerd:
Looks great, thank you!

PR #91, opened 2018-11-19
Handled malformed Content-Type params

dcormier:
Handled an issue when a content-type header like this is encountered:

`Content-Type: text/html;iso-8859-1`

### DIFF
--- a/detect_test.go
+++ b/detect_test.go
@@ -98,6 +98,16 @@ func TestDetectAttachmentHeader(t *testing.T) {
 				"Content-Disposition": []string{"inline"}},
 		},
 		{
+			want: false,
+			header: textproto.MIMEHeader{
+				"Content-Disposition": []string{"inline; broken"}},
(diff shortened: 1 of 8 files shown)


detect_test.go
```diff
@@ -97,6 +97,11 @@ func TestDetectAttachmentHeader(t *testing.T) {
 			header: textproto.MIMEHeader{
 				"Content-Disposition": []string{"inline"}},
 		},
+		{
+			want: false,
+			header: textproto.MIMEHeader{
+				"Content-Disposition": []string{"inline; broken"}},
+		},
```
jhillyerd: Add another entry to test `"attachment; broken"`

dcormier: Done.

jhillyerd:
Thanks again!

dcormier:
Gladly. Thanks for maintaining this!

PR #95, opened 2018-11-21
Wrapping errors with errors.WithStack()

### LINKED ISSUE #94: Consider wrapping returned errors with stack
When bugs are encountered, it's easier to track them down if there's a stack trace that goes with the error.

To that end, I know I would find it useful to wrap returned errors with [errors.WithStack](https://godoc.org/github.com/pkg/errors#WithStack).

As you can see from the two examples there, normal usage of the wrapped error will change nothing. But using the `%+v` format specifier will produce a useful stack trace.

Already when I'm debugging this library I sprinkle `errors.WithStack` in place to help me see what's going on, but I haven't (yet) included those changes in a PR. I think it would be a useful change for contributors to be able to use these stack traces to help track down bugs.

dcormier:
Closes #94.

### DIFF
--- a/boundary.go
+++ b/boundary.go
@@ -3,9 +3,10 @@ package enmime
 import (
 	"bufio"
 	"bytes"
-	"fmt"
 	"io"
 	"io/ioutil"
+
+	"github.com/pkg/errors"
 )
 
 // This constant needs to be at least 76 for this package to work correctly.  This is because
@@ -46,7 +47,7 @@ func (b *boundaryReader) Read(dest []byte) (n int, err error) {
 	peekEOF := (err == io.EOF)
 	if err != nil && !peekEOF && err != bufio.ErrBufferFull {
 		// Unexpected error
-		return 0, err
+		return 0, errors.WithStack(err)
 	}
 	var nCopy int
 	idx, complete := locateBoundary(peek, b.nlPrefix)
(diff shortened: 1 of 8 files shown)


dcormier:
Bah. Committed something I didn't mean to, there. I'll clean it up on Friday, probably.

dcormier:
There we go.

boundary.go
```diff
@@ -117,7 +120,7 @@ func (b *boundaryReader) Next() (bool, error) {
 			continue
 		}
 		b.finished = true
-		return false, fmt.Errorf("expecting boundary %q, got %q", string(b.prefix), string(line))
+		return false, errors.WithStack(fmt.Errorf("expecting boundary %q, got %q", string(b.prefix), string(line)))
```
jhillyerd:
```suggestion
		return false, errors.Errorf("expecting boundary %q, got %q", string(b.prefix), string(line))
```

looks like Errorf grabs stack: https://github.com/pkg/errors/blob/059132a15dd08d6704c67711dae0cf35ab991756/errors.go#L112

part.go
```diff
@@ -334,18 +335,18 @@ func parseParts(parent *Part, reader *bufio.Reader) error {
...
-				if err == io.EOF || strings.HasSuffix(err.Error(), "EOF") {
+				if errors.Cause(err) == io.EOF || strings.HasSuffix(err.Error(), "EOF") {
 					// There are no more Parts. The error must belong to the parent, because this
 					// part doesn't exist.
 					parent.addWarning(ErrorMissingBoundary, "Boundary %q was not closed correctly",
 						parent.Boundary)
 					break
 				}
-				return fmt.Errorf("error at boundary %v: %v", parent.Boundary, err)
+				return errors.WithMessage(err, fmt.Sprintf("error at boundary %v", parent.Boundary))
```
jhillyerd:
```suggestion
				return errors.Wrapf(err, "error at boundary %v", parent.Boundary)
```

dcormier: Added a comment explaing why `errors.WithMessage` is used there.

jhillyerd: Makes sense, we should probably at use WithMessagef then.

jhillyerd: ah, see the comment now, nvm.

PR #98, opened 2018-11-27
Better character set detection behavior

dcormier:
Only attempt to detect character set on `text/*` parts

On failure to detect character set, fall back to part-specified character set (which is what was used prior to charset detection being introduced) and add a warning to the part rather than fail to parse the message.

### DIFF
--- a/part.go
+++ b/part.go
@@ -155,6 +155,97 @@ func (p *Part) setupContentHeaders(mediaParams map[string]string) {
 	}
 }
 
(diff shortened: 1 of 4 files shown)


dcormier:
Related to issue #81 and the resolution of that issue in PR #87.

jhillyerd:
Nice, this certainly makes things easier to understand.  Thanks

dcormier:
Glad I can help.

PR #102, opened 2018-11-29
Remove CTE header if set in a part before encoding it.

kevinmahaffey:
In the case that a pre-existing part is decoded (e.g. from a pre-existing email), the content in an encoded part will be decoded correctly and the CTE header will be set based on the original part (e.g. quoted-printable). When encoding that part in the case that enmime's CTE detection decides that it should be 7bit, it did not remove the previous CTE header, resulting in a header specifying quoted-printable, but data that was encoded as 7bit.


jhillyerd:
Nice catch, and thank you for including a unit test!

I edited this to merge against `develop` instead of master; as master is for stable releases.  There is some info about this in the contributing file, but I really need to add it to the readme.

kevinmahaffey:
TDD for life :)

Ah, sorry I missed master vs. develop for releases.

–K

Issue #99, opened 2018-11-29
Part.Encode() adds an extra newline


dann7387:
What I did:
- Place string into a "text/plain" Part.
- Use Part.Encode() to write object to buffer
- Use Part.ReadParts() to read the buffer back into a Part object
- Read the Content field of the new object

```
package main

import (
        "bytes"
        "fmt"
        "github.com/jhillyerd/enmime"
)

func main() {
        var buf bytes.Buffer
        text := "This is my message."
        p := enmime.NewPart("text/plain")
        p.Content = []byte(text)

        p.Encode(&buf)
        p1, _ := enmime.ReadParts(&buf)

        fmt.Printf("\"%s\"", string(p1.Content))
}
```

What I expected:
Expected the resulting string to be exactly the same as the original.
```
"This is my message."
```

What I got:
Extra new line (\r\n) at the end of the string
```
"This is my message.
"
```

Release or branch I am using:
master

(Please attach a sample message if you feel it will help reproduce the issue)

jhillyerd:
https://travis-ci.org/jhillyerd/enmime/builds/461121525

Build broke on travis.

jhillyerd:
Ah, I think it's because the `testdata/encode/part-file-mod-date.golden` was added in a different commit.  will fix now.

PR #103, opened 2018-11-29
Handle non-delimited address lists

requaos:


requaos:
I tried to follow the TravisCI output, made no sense, couldn't find the test it mentioned and the line number in `encode_test.go:179` is just `if err != nil {`... fishy
```✔ ~/go/src/github.com/jhillyerd/enmime [requaos/multipleFrom|⚑ 1] 
18:08 $ make
go clean github.com/jhillyerd/enmime github.com/jhillyerd/enmime/cmd github.com/jhillyerd/enmime/cmd/mime-dump github.com/jhillyerd/enmime/cmd/mime-extractor github.com/jhillyerd/enmime/internal/coding github.com/jhillyerd/enmime/internal/stringutil github.com/jhillyerd/enmime/internal/test
go test -race ./...
ok      github.com/jhillyerd/enmime     2.162s
ok      github.com/jhillyerd/enmime/cmd 1.583s
?       github.com/jhillyerd/enmime/cmd/mime-dump       [no test files]
?       github.com/jhillyerd/enmime/cmd/mime-extractor  [no test files]
ok      github.com/jhillyerd/enmime/internal/coding     1.566s
ok      github.com/jhillyerd/enmime/internal/stringutil 1.553s
ok      github.com/jhillyerd/enmime/internal/test       1.566s
go build
```

jhillyerd:
Not your fault, I moved #102 from master to develop, so its golden test file does not reflect the newline changes from #99.  I will fix it tonight and re-run your test.

envelope_test.go
```diff
@@ -869,6 +869,24 @@ func TestUnquotedSpecialCharParamsInMime(t *testing.T) {
 	}
 }
 
+func TestBadAddressHeaderInMime(t *testing.T) {
+	msg := test.OpenTestData("mail", "malformedMultipleAddressHeaderValues.raw")
```
jhillyerd:
```suggestion
	msg := test.OpenTestData("mail", "malformed-multiple-address-header-values.raw")
```
All the other testdata files are kabob-case.  Obv need to rename the actualy file.

requaos: done

jhillyerd:
Thank you!

requaos:
As a follow up to this, since I was neck deep in the RFC yesterday, do we enforce this rule during encoding:
If multiple addresses are present in the "From" address header, then a single address in the "Sender" address header must exist.

Conversely, as long as only one "From" address is present, then the "Sender" header is not required.

Should I open an issue? Since I have not even investigated if this logic is present (TLDR)

jhillyerd:
Worth tracking in an issue.  I believe that the builder does not allow multiple From (or any Sender); and its build method is where most validation happens at this point, which is not much at all.

My feeling is if someone wants to build an email with Part objects manually, then they need to do their own validation.

Issue #112, opened 2018-12-18
Errors in address format processing without white space


derktam:
**What I did:**
```
...
From: "=?UTF-8?B?7ZWc6rWt642w7J207YSw7KeE7Z2l7JuQ?="<kodb_pr@kdata.or.kr>
...
```
The following problem occurred during testing with the eml file containing the above sender.

When testing with the same data,  
using the ***mail*** package's **mail.ParseAddressList(m.GetHeader("From"))** works correctly,  
but when using **Envelope.AddressList("From")**, a **mail: no angle-addr** error occurs.

**What I expected:**
([]\*Mail.Address) returned without error,

**What I got:**
returned error
**mail: no angle-addr**

**Release or branch I am using:**
both v0.5.0 and master

(Please attach a sample message if you feel it will help reproduce the issue)

derktam:
The address part(<kodb_pr@kdata.or.kr>) disappears after passing through decodeToUTF8Base64Header() in AddressList().

jhillyerd:
Thanks, definitely sounds like a bug.

requaos:
I'll take a look today

jhillyerd:
If easy, we may also want to add a warning to the Errors list.  Not essential.

requaos:
@derktam Could you please confirm if your issue is now resolved using the latest develop branch?

derktam:
It seems to have been resolved. Thank you :)

requaos:
👍

PR #113, opened 2019-01-21
Fix for #112

requaos:
In this PR I am also proposing to make our "ParseMediaType" function available in the public api. In my project I have needed this for a thin stdlib pre-parser, to distinguish if I should even bother sending the bytes to enmime. The kind of realworld nuances we have navigated in this function, and the ones we have yet to navigate, make me want to avoid copy 'n paste solution.

### DIFF
--- a/detect.go
+++ b/detect.go
@@ -9,7 +9,7 @@ import (
 func detectMultipartMessage(root *Part) bool {
 	// Parse top-level multipart
 	ctype := root.Header.Get(hnContentType)
-	mediatype, _, _, err := parseMediaType(ctype)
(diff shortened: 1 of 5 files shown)


jhillyerd:
I'm not opposed to making ParseMediaType public, your reasoning makes sense to me.

I don't think we should do it in the same PR as an unrelated change though.

requaos:
I will split it out then, thank you

requaos:
PR split into https://github.com/jhillyerd/enmime/pull/114 and https://github.com/jhillyerd/enmime/pull/115 respectively

PR #118, opened 2019-01-24
Touchup of RCF2047 mime attribute parameter decoding

requaos:


header_test.go
```diff
@@ -194,6 +194,11 @@ func TestFixMangledMediaType(t *testing.T) {
 			sep:   ";",
 			want:  "one/two; name=\"file.two\"",
 		},
+		{
+			input: "application/octet-stream; =?UTF-8?B?bmFtZT3DsMKfwpTCii5tc2c=?=",
+			sep:   " ",
+			want:  "application/octet-stream;name=\"ð.msg",
```
jhillyerd: Shouldn't this have a trailing `\"`?

jhillyerd: Sounds good. :)

PR #121, opened 2019-01-25
Implementation which does not depend on successive function calls

requaos:
Only quoting the head of the attribute value and leaving the end unquoted was a hack. This will now properly tokenize the string, handle multiple RFC2047 chunks if present and quote both ends of the attribute value as needed.

### DIFF
--- a/header.go
+++ b/header.go
@@ -521,40 +521,40 @@ func whiteSpaceRune(r rune) bool {
 // rfc2047AttributeName checks if the attribute name is encoded in RFC2047 format
 // RFC2047 Example:
 //     `=?UTF-8?B?bmFtZT0iw7DCn8KUwoo=?=`
-func rfc2047AttributeName(s string) string {
-	if !strings.Contains(s, "?=") {
-		return s
+func rfc2047AttributeName(name string) string {
+	if !strings.Contains(name, "?=") {
+		return name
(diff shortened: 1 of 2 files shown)


requaos:
Force-pushed for rebase on develop

header.go
```diff
@@ -509,16 +525,47 @@ func rfc2047AttributeName(s string) string {
...
-	pair := strings.SplitAfter(s, "?=")
-	// lets assume that the attribute was encoded because of unicode characters being present
-	// then the attribute value should be quoted
-	keyValuePair := strings.SplitAfter(decodeHeader(pair[0]), "=")
-	// only quote the parameter value if it isn't already quoted
-	if len(keyValuePair) > 1 {
-		if !strings.HasPrefix(keyValuePair[1], "\"") {
-			keyValuePair[1] = fmt.Sprintf("\"%s", keyValuePair[1])
+	// copy the string so we can return the original if we encounter any issues
+	temp := s
```
jhillyerd: `temp` is a code smell.  I think what I'd do here is replace`s` with `name` entirely in this function, and rename `temp` to `s`.  What do you think?

requaos: I like it. Done.

jhillyerd:
Thanks!

PR #123, opened 2019-01-30
Requaos/nontermboundry

requaos:
Had an issue where the boundary was not terminated, and I saw some tests around this but my case fell through the cracks. I refactored the `Read()` function to go one byte at a time and only peek the length of the `fullBoundary`.
(edited)
There was a case where the boundary was being detected 2x chars early for whitespace at the head, resolved in b6a4d4e

### DIFF
--- a/boundary.go
+++ b/boundary.go
@@ -76,13 +76,13 @@ func (b *boundaryReader) Read(dest []byte) (n int, err error) {
 	}
 
 	for i := 0; i < cap(dest); i++ {
-		c, err := b.r.Peek(1)
+		cs, err := b.r.Peek(1)
 		if err != nil && err != io.EOF {
 			return 0, errors.WithStack(err)
 		}
-		// Ensure that we can switch on the first byte of 'c' without panic.
-		if len(c) > 0 {
-			switch c[0] {
+		// Ensure that we can switch on the first byte of 'cs' without panic.
+		if len(cs) > 0 {
+			switch cs[0] {
 			// Check for line feed as potential LF boundary prefix.
 			case '\n':
 				peek, err := b.r.Peek(len(b.nlPrefix) + 2)
--- a/boundary.go
+++ b/boundary.go
@@ -20,15 +20,15 @@ const peekBufferSize = 4096
 var errNoBoundaryTerminator = stderrors.New("expected boundary not present")
 
 type boundaryReader struct {
-	finished        bool          // No parts remain when finished
-	partsRead       int           // Number of parts read thus far
-	r               *bufio.Reader // Source reader
-	nlPrefix        []byte        // NL + MIME boundary prefix
-	prefix          []byte        // MIME boundary prefix
-	final           []byte        // Final boundary prefix
-	buffer          *bytes.Buffer // Content waiting to be read
-	crBoundryPrefix bool          // Flag for CR in CRLF + MIME boundary
-	unbounded       bool          // Flag to throw errNoBoundaryTerminator
+	finished         bool          // No parts remain when finished
+	partsRead        int           // Number of parts read thus far
+	r                *bufio.Reader // Source reader
+	nlPrefix         []byte        // NL + MIME boundary prefix
+	prefix           []byte        // MIME boundary prefix
+	final            []byte        // Final boundary prefix
+	buffer           *bytes.Buffer // Content waiting to be read
+	crBoundaryPrefix bool          // Flag for CR in CRLF + MIME boundary
+	unbounded        bool          // Flag to throw errNoBoundaryTerminator
 }
 
 // newBoundaryReader returns an initialized boundaryReader
@@ -67,13 +67,16 @@ func (b *boundaryReader) Read(dest []byte) (n int, err error) {
 				// check for a match against boundary or terminal boundary
 				if b.isDelimiter(peek[1:]) || b.isTerminator(peek[1:]) {
 					// if we stashed a carriage return, lets pop that back onto the io.Reader
-					if b.crBoundryPrefix {
+					if b.crBoundaryPrefix {
 						err = b.r.UnreadByte()
-						if err != nil {
+						switch err {
+						case nil, bufio.ErrInvalidUnreadByte:
+							// carriage return boundary prefix bit already unread
+						default:
 							// this should never happen, fatal error
 							return 0, errors.WithStack(err)
 						}
-						b.crBoundryPrefix = false
+						b.crBoundaryPrefix = false
 					}
 					n, err = b.buffer.Read(dest)
 					switch err {
@@ -92,12 +95,12 @@ func (b *boundaryReader) Read(dest []byte) (n int, err error) {
 				continue
 			}
 			// wasn't a boundary, write the stored carriage return
-			if b.crBoundryPrefix {
+			if b.crBoundaryPrefix {
 				err = b.buffer.WriteByte(byte('\r'))
 				if err != nil {
 					return 0, errors.WithStack(err)
 				}
-				b.crBoundryPrefix = false
+				b.crBoundaryPrefix = false
 			}
 		}
 
@@ -107,7 +110,7 @@ func (b *boundaryReader) Read(dest []byte) (n int, err error) {
 			if err != nil {
 				return 0, errors.WithStack(err)
 			}
-			b.crBoundryPrefix = true
+			b.crBoundaryPrefix = true
 			continue
 		}
 
--- a/boundary.go
+++ b/boundary.go
@@ -3,6 +3,7 @@ package enmime
 import (
 	"bufio"
 	"bytes"
+	stderrors "errors"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -16,7 +17,7 @@ import (
 // from it.
 const peekBufferSize = 4096
 
-var errNoBoundaryTerminator = errors.New("expected boundary not present")
+var errNoBoundaryTerminator = stderrors.New("expected boundary not present")
 
 type boundaryReader struct {
 	finished        bool          // No parts remain when finished
--- a/boundary.go
+++ b/boundary.go
@@ -44,79 +44,118 @@ func newBoundaryReader(reader *bufio.Reader, boundary string) *boundaryReader {
 }
 
 // Read returns a buffer containing the content up until boundary
+//
+//   Excerpt from io package on io.Reader implementations:
+//
+//     type Reader interface {
+//        Read(p []byte) (n int, err error)
+//     }
+//
+//     Read reads up to len(p) bytes into p. It returns the number of
+//     bytes read (0 <= n <= len(p)) and any error encountered. Even
+//     if Read returns n < len(p), it may use all of p as scratch space
+//     during the call. If some data is available but not len(p) bytes,
+//     Read conventionally returns what is available instead of waiting
+//     for more.
+//
+//     When Read encounters an error or end-of-file condition after
+//     successfully reading n > 0 bytes, it returns the number of bytes
+//     read. It may return the (non-nil) error from the same call or
+//     return the error (and n == 0) from a subsequent call. An instance
+//     of this general case is that a Reader returning a non-zero number
+//     of bytes at the end of the input stream may return either err == EOF
+//     or err == nil. The next Read should return 0, EOF.
+//
+//     Callers should always process the n > 0 bytes returned before
+//     considering the error err. Doing so correctly handles I/O errors
+//     that happen after reading some bytes and also both of the allowed
+//     EOF behaviors.
 func (b *boundaryReader) Read(dest []byte) (n int, err error) {
 	if b.buffer.Len() >= len(dest) {
-		// This read request can be satisfied entirely by the buffer
+		// This read request can be satisfied entirely by the buffer.
 		return b.buffer.Read(dest)
 	}
 
 	for i := 0; i < cap(dest); i++ {
-		oneByte, err := b.r.Peek(1)
+		c, err := b.r.Peek(1)
 		if err != nil && err != io.EOF {
 			return 0, errors.WithStack(err)
 		}
-		if bytes.Equal(oneByte, []byte{'\n'}) {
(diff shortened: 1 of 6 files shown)


requaos:
Gonna brush up the code coverage a little. In the meantime, let me know your thoughts around this change.

requaos:
Still working on this one, please don't merge yet

requaos:
So, after looking at the actual issue I was having. The problem I am trying to solve here is that the part has no boundary terminator, which during parsing was yielding unexpected EOF.

> 1. Can we create a simulation of your failing case for a unit test?
Yes, I'm pushing that now
> 2. ...It would be good to have before/after benchmarks...
I agree. When running current `develop` with `go test -a -count=100 ./...` versus this branch, there does not seem to be much difference. The underlying io.Reader is a bufio.Reader, which already chunks the stream into the buffer. Another point here is that in this `Read()` implementation, can only write `512`bytes to the destination slice per iteration, so allocating `4`kb on each run felt like overkill.

I can write some benchmarks, but that will take some time when I can spin everything down on my machine for accurate values.

A note about the code coverage: I noticed that coveralls shows a few files at 0.0% coverage while `goconvey` (I just use the web UI and not the framework) shows it at 100%... One in particular is the hand-rolled UUIDv4 implementation. We could drop that from the repo and use the one from [gofrs/uuid](https://github.com/gofrs/uuid) to remove a 0.0% from the average.

requaos:
At the end this refactor may be unwarranted, as the result I am attempting to elicit is still not achieved. My goal in this was that if the terminal boundary is not present that the boundary reader would still return up to the end as the part and add the `Missing Boundary` warning to the errors slice.
What I get is the warning on the error slice, but the part never gets appended. I still want the content, even if it's partial

requaos:
Objective achieved. All the tests pass and we now get the, potentially incomplete, content on the part in addition to the warning on the errors slice

jhillyerd:
Starting to read this.  I'm going to have a lot of little nit-picks, apologizing in advance if this takes a few rounds of reviewing. :)

boundary.go
```diff
@@ -43,42 +50,81 @@ func (b *boundaryReader) Read(dest []byte) (n int, err error) {
...
-		if nCopy = len(peek) - len(b.nlPrefix) - 1; nCopy <= 0 {
-			nCopy = 0
-			if peekEOF {
-				// No more peek space remaining and no boundary found
-				return 0, errors.WithStack(io.ErrUnexpectedEOF)
+		if bytes.Equal(oneByte, []byte{'\n'}) {
+			peek, err := b.r.Peek(len(b.nlPrefix) + 2)
+			switch err {
+			case nil:
+				// don't check the boundary until we only have one new line
```
jhillyerd: Go comments should be a full sentence, starting with capital (unless variable name) and ending with period.  enmime is still a work on progress on this front.  Please update other comments below.

boundary.go
```diff
@@ -43,42 +50,81 @@ func (b *boundaryReader) Read(dest []byte) (n int, err error) {
...
+						switch err {
+						case nil, bufio.ErrInvalidUnreadByte:
+							// carriage return boundary prefix bit already unread
+						default:
+							// this should never happen, fatal error
+							return 0, errors.WithStack(err)
+						}
+						b.crBoundaryPrefix = false
+					}
+					n, err = b.buffer.Read(dest)
```
jhillyerd: I think a comment above this line explaining that we've found a boundary and are copying the preceding content or similar would be nice for the reader to understand our current state.

boundary.go
```diff
@@ -43,42 +50,81 @@ func (b *boundaryReader) Read(dest []byte) (n int, err error) {
...
+					switch err {
+					case nil, io.EOF:
+						return n, io.EOF
+					default:
+						return 0, errors.WithStack(err)
+					}
+				}
+			default:
+				// got to the end without seeing a boundary
+				if err == io.EOF {
```
jhillyerd: I don't understand why we are in a `switch err` but then testing err for equality, can we pull this out into a `io.EOF:` block?

boundary.go
```diff
@@ -43,42 +50,81 @@ func (b *boundaryReader) Read(dest []byte) (n int, err error) {
...
+					b.unbounded = true
+					break
+				}
+				continue
+			}
+			// wasn't a boundary, write the stored carriage return
+			if b.crBoundaryPrefix {
+				err = b.buffer.WriteByte(byte('\r'))
```
jhillyerd: Isn't the cast to byte redundant?

boundary.go
```diff
@@ -43,42 +50,81 @@ func (b *boundaryReader) Read(dest []byte) (n int, err error) {
...
+				b.crBoundaryPrefix = false
 			}
 		}
-	}
-	if nCopy > 0 {
-		if _, err = io.CopyN(b.buffer, b.r, int64(nCopy)); err != nil {
-			return 0, errors.WithStack(err)
+
+		// store this carriage return just in case it begins a boundary
+		if bytes.Equal(oneByte, []byte{'\r'}) {
```
jhillyerd:
```suggestion
		if c == '\r' {
```

boundary.go
```diff
@@ -119,7 +165,7 @@ func (b *boundaryReader) Next() (bool, error) {
 			continue
 		}
 		b.finished = true
-		return false, errors.Errorf("expecting boundary %q, got %q", string(b.prefix), string(line))
+		return false, errors.WithMessage(errNoBoundaryTerminator, fmt.Sprintf("expecting boundary %q, got %q", string(b.prefix), string(line)))
```
jhillyerd:
```suggestion
		return false, errors.WithMessagef(errNoBoundaryTerminator, "expecting boundary %q, got %q", string(b.prefix), string(line))
```

boundary.go
```diff
@@ -132,9 +178,8 @@ func (b *boundaryReader) isDelimiter(buf []byte) bool {
 
 	// Fast forward to the end of the boundary prefix
 	buf = buf[idx+len(b.prefix):]
-	buf = bytes.TrimLeft(buf, " \t")
 	if len(buf) > 0 {
-		if buf[0] == '\r' || buf[0] == '\n' {
+		if unicode.IsSpace(rune(buf[0])) {
```
jhillyerd: Do we have already have test coverage for this change?

boundary_test.go
```diff
@@ -290,8 +290,8 @@ func TestBoundaryReaderNoTerminator(t *testing.T) {
 		t.Fatal("Next() = false, want: true")
 	}
 
-	// Second part should error
-	want := "expecting boundary"
+	// There is no second part should, error should be EOF
```
jhillyerd:
```suggestion
	// There is no second part should, error should be EOF.
```

boundary.go
```diff
@@ -44,86 +44,126 @@ func newBoundaryReader(reader *bufio.Reader, boundary string) *boundaryReader {
...
 func (b *boundaryReader) Read(dest []byte) (n int, err error) {
 	if b.buffer.Len() >= len(dest) {
-		// This read request can be satisfied entirely by the buffer
+		// This read request can be satisfied entirely by the buffer.
 		return b.buffer.Read(dest)
 	}
 
 	for i := 0; i < cap(dest); i++ {
-		oneByte, err := b.r.Peek(1)
+		c, err := b.r.Peek(1)
```
jhillyerd: Ah, sorry, the `oneByte` confused me, thought it was a byte instead of a slice.  `cs` would be more idiomatic for a slice of bytes.  My bad.

boundary.go
```diff
@@ -37,48 +43,127 @@ func newBoundaryReader(reader *bufio.Reader, boundary string) *boundaryReader {
...
+			return 0, errors.WithStack(err)
 		}
-	} else {
-		// No boundary found, move forward a safe distance
-		if nCopy = len(peek) - len(b.nlPrefix) - 1; nCopy <= 0 {
-			nCopy = 0
-			if peekEOF {
-				// No more peek space remaining and no boundary found
-				return 0, errors.WithStack(io.ErrUnexpectedEOF)
+		// Ensure that we can switch on the first byte of 'c' without panic
```
jhillyerd:
```suggestion
		// Ensure that we can switch on the first byte of 'c' without panic.
```

boundary.go
```diff
@@ -37,48 +43,127 @@ func newBoundaryReader(reader *bufio.Reader, boundary string) *boundaryReader {
...
+		if len(c) > 0 {
+			switch c[0] {
+			// Check for line feed as potential LF boundary prefix
```
jhillyerd:
```suggestion
			// Check for line feed as potential LF boundary prefix.
```

boundary.go
```diff
@@ -37,48 +43,127 @@ func newBoundaryReader(reader *bufio.Reader, boundary string) *boundaryReader {
...
+						// Check if we stored a carriage return.
+						if b.crBoundaryPrefix {
+							b.crBoundaryPrefix = false
+							// Let us now unread that back onto the io.Reader, since
+							// we have found what we are looking for and this byte
+							// belongs to the bounded block we are reading.
+							err = b.r.UnreadByte()
+							switch err {
+							case nil:
+								// Carry on
```
jhillyerd:
```suggestion
								// Carry on.
```

boundary.go
```diff
@@ -37,48 +43,127 @@ func newBoundaryReader(reader *bufio.Reader, boundary string) *boundaryReader {
...
+							case bufio.ErrInvalidUnreadByte:
+								// Carriage return boundary prefix bit already unread
```
jhillyerd:
```suggestion
								// Carriage return boundary prefix bit already unread.
```

boundary.go
```diff
@@ -37,48 +43,127 @@ func newBoundaryReader(reader *bufio.Reader, boundary string) *boundaryReader {
...
+				// Checked '\n' was not prefix to a boundary.
+				if b.crBoundaryPrefix {
+					b.crBoundaryPrefix = false
+					// Stored '\r' should be written to the buffer now.
+					err = b.buffer.WriteByte('\r')
+					if err != nil {
+						return 0, errors.WithStack(err)
+					}
+				}
+			// Check for carriage return as potential CRLF boundary prefix
```
jhillyerd:
```suggestion
			// Check for carriage return as potential CRLF boundary prefix.
```

boundary.go
```diff
@@ -37,48 +43,127 @@ func newBoundaryReader(reader *bufio.Reader, boundary string) *boundaryReader {
...
+			case '\r':
+				_, err := b.r.ReadByte()
+				if err != nil {
+					return 0, errors.WithStack(err)
+				}
+				// Flag the boundary reader to indicate that we
+				// have stored a '\r' as a potential CRLF prefix
```
jhillyerd:
```suggestion
				// have stored a '\r' as a potential CRLF prefix.
```

boundary.go
requaos: I am adding a test case for this change to TestBoundaryReader:
```
		{
			input:    "good\r\n--STOPHERE\t\r\nafter",
			boundary: "STOPHERE",
			want:     "good",
		},
```
Found use case in the wild.

boundary.go
```diff
@@ -37,48 +43,127 @@ func newBoundaryReader(reader *bufio.Reader, boundary string) *boundaryReader {
...
 		}
-	}
-	if nCopy > 0 {
-		if _, err = io.CopyN(b.buffer, b.r, int64(nCopy)); err != nil {
-			return 0, errors.WithStack(err)
+
+		_, err = io.CopyN(b.buffer, b.r, 1)
+		if err != nil {
+			// EOF is not fatal, it just means
+			// that we have drained the reader.
```
jhillyerd: I think this comment can fit on one line.

requaos:
@jhillyerd, It looks like the Go-Modules release of `github.com/pkg/errors` has yet to include `errors.WithMessagef` in it's API...  I realize now thats why I was using `errors.WithMessage` and `fmt.Sprintf` in conjunction.

requaos:
As an aside, I am still tracing one intermittent issue with this refactor.
There is a scenario where immediately after an initial boundary, as it is written to the buffer during `setupHeaders` read call, the trailing `\n` in `--someBoundary100\r\n` is omitted/dropped. This becomes evident during a `Next()` call on the boundary reader when `ReadLine('\n')` is called. The bytes returned run past the `\r` at the end of the boundary and include the first header line in that read. This causes the underlying reader to move past the first header in the MIME block.

jhillyerd:
Acknowledged, won't merge until you say it's ready.  If you figure out a consistent repro for the intermittent issue, maybe add a unit test for it.

requaos:
I had a test setup for reproduction, but it is fixed now and not reproducible. Two changes worth noting that may have fixed the problem:

- The check for `nil` on slice of `c`
- Moving the `b.crBoundaryPrefix = false` state setting line after the check, before any logic (This is the more likely of the two)

The real-world email that I was testing with to reproduce this, now parses fine.

I believe the last change request remaining is to rename `c` (Character) to `cs` (Character Slice)

boundary.go
requaos: Done

boundary.go
requaos: Done

boundary.go
requaos: Done

boundary.go
requaos: Done

boundary.go
requaos: Done

boundary.go
requaos: Done

boundary.go
requaos: Done

boundary.go
requaos: Done

boundary_test.go
requaos: Done

boundary.go
requaos: I added this test case

requaos:
On my side this is ready, but please keep the comments coming. I need more practice using idiomatic conventions.

jhillyerd:
Everything looks good to me.  Thanks for all the hard work on this!

PR #124, opened 2019-02-05
media parameter attribute value needs to escape quotes inside quoted …

requaos:
…value

### DIFF
--- a/header.go
+++ b/header.go
@@ -235,8 +235,9 @@ func ParseMediaType(ctype string) (mtype string, params map[string]string, inval
 			}
 			mtype, params, err = mime.ParseMediaType(mctype)
 			if err != nil {
-				// If the media parameter has special characters, ensure that it is quoted.
-				mtype, params, err = mime.ParseMediaType(fixUnquotedSpecials(mctype))
+				// If the media parameter has special characters, ensure that
+				// it is quoted and that any existing quotes are escaped.
+				mtype, params, err = mime.ParseMediaType(fixUnescapedQuotes(fixUnquotedSpecials(mctype)))
 				if err != nil {
 					return "", nil, nil, errors.WithStack(err)
 				}
@@ -511,6 +512,78 @@ func fixUnquotedSpecials(s string) string {
 	}
 
 	return clean.String()
+}
+
+func fixUnescapedQuotes(s string) string {
+	split := strings.SplitAfter(s, ";")
+	fresh := &strings.Builder{}
+	for i := 0; i < len(split); i++ {
+		// find the "="
+		idx := strings.IndexByte(split[i], '=')
+		if idx < 0 {
+			// no "=", must be the content-type or a comment
+			fresh.WriteString(split[i])
+			continue
+		}
+		fresh.WriteString(split[i][:idx])
+		param := split[i][idx:]
+		openingQuoteIdx := strings.IndexByte(param, '"')
+		closingQuoteIdx := strings.LastIndexByte(param, '"')
+		// cut out early if there are no quotes
+		if openingQuoteIdx < 0 && closingQuoteIdx < 0 {
+			// value isn't quoted
+			fresh.WriteString(param)
+			continue
+		}
+		// lets append the next chunk of split here in case of a semicolon mid string
+		if closingQuoteIdx == openingQuoteIdx {
+			param = fmt.Sprintf("%s%s", param, split[i+1])
+			closingQuoteIdx = strings.LastIndexByte(param, '"')
+			i++
+		}
+		// it's got quotes, lets put the k/v separator back in along with everything upto the first quote
+		fresh.WriteByte('=')
+		// quotes start here
+		fresh.WriteByte('"')
+		fresh.WriteString(param[1:openingQuoteIdx])
+		// get just the value, less the outer quotes
+		rest := param[closingQuoteIdx+1:]
+		// if there is stuff after the last quote then we should escape the first quote
+		if len(rest) > 0 {
+			fresh.WriteString("\\\"")
+		}
+		param = param[openingQuoteIdx+1 : closingQuoteIdx]
+		alreadyEscaped := false
+		for strIdx := range param {
+			switch param[strIdx] {
+			case '"':
+				// we are inside of a quoted string, so lets escape this guy if it isn't already escaped
+				if !alreadyEscaped {
+					fresh.WriteByte('\\')
+					alreadyEscaped = false
+				}
+				fresh.WriteByte(param[strIdx])
+			case '\\':
+				// something is getting escaped, a quote is the only char that needs
+				// this, so lets assume the following char is a double-quote
+				alreadyEscaped = true
+				fresh.WriteByte('\\')
+			default:
+				fresh.WriteByte(param[strIdx])
+			}
+		}
+		// if there is stuff after the last quote then we should escape
+		// the last quote, apply the rest and terminate with a quote
+		if len(rest) > 0 {
+			fresh.WriteByte('\\')
+			fresh.WriteByte('"')
+			fresh.WriteString(rest)
+			fresh.WriteByte('"')
+		} else {
+			fresh.WriteByte('"')
+		}
+	}
+	return fresh.String()
 }
 
 // Detects a RFC-822 linear-white-space, passed to strings.FieldsFunc.
--- a/header.go
+++ b/header.go
@@ -228,16 +228,16 @@ func ParseMediaType(ctype string) (mtype string, params map[string]string, inval
 		mctype := fixMangledMediaType(ctype, ";")
 		mtype, params, err = mime.ParseMediaType(mctype)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


header.go
```diff
@@ -513,6 +514,85 @@ func fixUnquotedSpecials(s string) string {
 	return clean.String()
 }
```
jhillyerd: Could you add a very short description of func with an example so the reader can visualize the transformation it's making?

header.go
```diff
@@ -513,6 +514,85 @@ func fixUnquotedSpecials(s string) string {
...
 
+func fixUnescapedQuotes(s string) string {
+	split := strings.SplitAfter(s, ";")
+	fresh := &strings.Builder{}
```
jhillyerd:
```suggestion
	sb:= &strings.Builder{}
```

for String Builder; although `fresh` did make me smile.

header.go
```diff
@@ -513,6 +514,85 @@ func fixUnquotedSpecials(s string) string {
...
+	for i := 0; i < len(split); i++ {
+		// find the "="
+		idx := strings.IndexByte(split[i], '=')
+		if idx < 0 {
+			// no "=", must be the content-type or a comment
```
jhillyerd:
```suggestion
			// No "=", must be the content-type or a comment.
```

header.go
```diff
@@ -513,6 +514,85 @@ func fixUnquotedSpecials(s string) string {
 	return clean.String()
 }
 
+func fixUnescapedQuotes(s string) string {
+	split := strings.SplitAfter(s, ";")
+	fresh := &strings.Builder{}
+	for i := 0; i < len(split); i++ {
+		// find the "="
+		idx := strings.IndexByte(split[i], '=')
```
jhillyerd:
```suggestion
		eq := strings.IndexByte(split[i], '=')
```

I don't see `idx` used in Go; for example, it's not used once in the entire strings pkg.  I do think that the Go authors sometimes overuse the single character variables which can be confusing in longer functions.

header.go
```diff
@@ -513,6 +514,85 @@ func fixUnquotedSpecials(s string) string {
...
+		if idx < 0 {
+			// no "=", must be the content-type or a comment
+			fresh.WriteString(split[i])
+			continue
+		}
+		fresh.WriteString(split[i][:idx])
+		param := split[i][idx:]
+		openingQuoteIdx := strings.IndexByte(param, '"')
```
jhillyerd:
```suggestion
		openingQuote := strings.IndexByte(param, '"')
```

header.go
```diff
@@ -513,6 +514,85 @@ func fixUnquotedSpecials(s string) string {
...
+		closingQuoteIdx := strings.LastIndexByte(param, '"')
```
jhillyerd:
```suggestion
		closingQuote := strings.LastIndexByte(param, '"')
```

header.go
```diff
@@ -513,6 +514,85 @@ func fixUnquotedSpecials(s string) string {
...
+		// cut out early if there are no quotes
```
jhillyerd:
```suggestion
		// Cut out early if there are no quotes.
```

header.go
```diff
@@ -513,6 +514,85 @@ func fixUnquotedSpecials(s string) string {
...
+		if openingQuoteIdx < 0 && closingQuoteIdx < 0 {
+			// value isn't quoted
```
jhillyerd:
```suggestion
			// Value isn't quoted.
```

header.go
```diff
@@ -513,6 +514,85 @@ func fixUnquotedSpecials(s string) string {
...
+			fresh.WriteString(param)
+			continue
+		}
+		// lets append the next chunk of split here in case of a semicolon mid string
```
jhillyerd:
```suggestion
		// Lets append the next chunk of split here in case of a semicolon mid string.
```

header.go
```diff
@@ -513,6 +514,85 @@ func fixUnquotedSpecials(s string) string {
...
+		if closingQuoteIdx == openingQuoteIdx {
+			param = fmt.Sprintf("%s%s", param, split[i+1])
+			closingQuoteIdx = strings.LastIndexByte(param, '"')
+			i++
+			if closingQuoteIdx == openingQuoteIdx {
+				return fresh.String()
+			}
+		}
+		// it's got quotes, lets put the k/v separator back in along with everything upto the first quote
```
jhillyerd:
```suggestion
		// It's got quotes, lets put the k/v separator back in along with everything up to the first quote.
```

header.go
```diff
@@ -513,6 +514,85 @@ func fixUnquotedSpecials(s string) string {
...
+		fresh.WriteByte('=')
+		// quotes start here
```
jhillyerd:
```suggestion
		// Quotes start here.
```

header.go
```diff
@@ -513,6 +514,85 @@ func fixUnquotedSpecials(s string) string {
...
+		fresh.WriteByte('"')
+		fresh.WriteString(param[1:openingQuoteIdx])
+		// get just the value, less the outer quotes
+		rest := param[closingQuoteIdx+1:]
+		// if there is stuff after the last quote then we should escape the first quote
```
jhillyerd:
```suggestion
		// If there is stuff after the last quote then we should escape the first quote.
```

header.go
```diff
@@ -513,6 +514,85 @@ func fixUnquotedSpecials(s string) string {
...
+		if len(rest) > 0 && rest != ";" {
+			fresh.WriteString("\\\"")
+		}
+		param = param[openingQuoteIdx+1 : closingQuoteIdx]
+		alreadyEscaped := false
+		for strIdx := range param {
```
jhillyerd: I don't feel like the name `strIdx` conveys anything unique, and this loop is fairly short.

```suggestion
		for j := range param {
```

header.go
```diff
@@ -513,6 +514,85 @@ func fixUnquotedSpecials(s string) string {
...
+			switch param[strIdx] {
+			case '"':
+				// we are inside of a quoted string, so lets escape this guy if it isn't already escaped
```
jhillyerd:
```suggestion
				// We are inside of a quoted string, so lets escape this guy if it isn't already escaped.
```

header.go
```diff
@@ -513,6 +514,85 @@ func fixUnquotedSpecials(s string) string {
...
+				if !alreadyEscaped {
+					fresh.WriteByte('\\')
+					alreadyEscaped = false
+				}
+				fresh.WriteByte(param[strIdx])
+			case '\\':
+				// something is getting escaped, a quote is the only char that needs
```
jhillyerd:
```suggestion
				// Something is getting escaped, a quote is the only char that needs
```

header.go
```diff
@@ -513,6 +514,85 @@ func fixUnquotedSpecials(s string) string {
...
+				// this, so lets assume the following char is a double-quote
```
jhillyerd:
```suggestion
				// this, so lets assume the following char is a double-quote.
```

header.go
```diff
@@ -513,6 +514,85 @@ func fixUnquotedSpecials(s string) string {
...
+				alreadyEscaped = true
+				fresh.WriteByte('\\')
+			default:
+				fresh.WriteByte(param[strIdx])
+			}
+		}
+		// if there is stuff after the last quote then we should escape
```
jhillyerd:
```suggestion
		// If there is stuff after the last quote then we should escape
```

header.go
```diff
@@ -513,6 +514,85 @@ func fixUnquotedSpecials(s string) string {
...
+		// the last quote, apply the rest and terminate with a quote
```
jhillyerd:
```suggestion
		// the last quote, apply the rest and terminate with a quote.
```

header.go
jhillyerd: I think GH got confused here, this was meant to go above the func fixUnescapedQuotes definition

header.go
```diff
@@ -513,6 +514,82 @@ func fixUnquotedSpecials(s string) string {
 	return clean.String()
 }
 
+func fixUnescapedQuotes(s string) string {
```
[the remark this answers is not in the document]
requaos: No, I totally agree. When I am visualizing how a stream is being read or how functions will interact, I just throw var names at the wall. Going back over it once functionality is achieved to apply Go Idioms is just a part of the development process. Thank you.

header.go
requaos: Huh, I must have carried that over from my C# or Java days, lol. Good point. I believe that shorter var names would make the intermediate AST smaller during compile time, but in practice I think it makes it easier to see the business logic abstracted away from the preconceptions that wordy var names can imply... I don't practice this enough, so I digress.

requaos:
Thank you for the feedback. I have applied your suggestions where appropriate and some I just reworded (with capitalization and punctuation in mind). And I fixed a bug related to `escaped` state where if it was not a `"` character following the `\` the flag would continue to indicate that an escape character had preceded it when it eventually encounters a `"` character.

requaos:
Please let me know if you have any additional feedback

PR #125, opened 2019-02-06
return empty string and nil error if content-type is empty so that en…

requaos:
…mime.ErrorMissingContentType can be added to the errors slice and continue parsing

### DIFF
--- a/part.go
+++ b/part.go
@@ -296,12 +296,13 @@ func (p *Part) decodeContent(r io.Reader) error {
 	// Collect base64 errors.
 	if b64cleaner != nil {
 		for _, err := range b64cleaner.Errors {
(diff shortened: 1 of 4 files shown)


part.go
```diff
@@ -296,13 +296,14 @@ func (p *Part) decodeContent(r io.Reader) error {
...
 		for _, err := range b64cleaner.Errors {
-			p.Errors = append(p.Errors, &Error{
-				Name:   ErrorMalformedBase64,
-				Detail: err.Error(),
-				Severe: false,
-			})
+			p.addWarning(ErrorMalformedBase64, err.Error())
 		}
 	}
+	// Set empty content-type error
```
jhillyerd:
```suggestion
	// Set empty content-type error.
```

PR #129, opened 2019-05-22
Propper detection of text/plain attachment.

davrux:
Mail with content headers

Content-Type: text/plain; name="test.csv"
Content-Disposition: attachment; filename="test.csv"

is parsed incorrectly. The CSV attachment is used as text body.

### DIFF
--- a/detect.go
+++ b/detect.go
@@ -63,6 +63,16 @@ func detectTextHeader(header textproto.MIMEHeader, emptyContentTypeIsText bool) 
 // detectBinaryBody returns true if the mail header defines a binary body.
 func detectBinaryBody(root *Part) bool {
 	if detectTextHeader(root.Header, true) {
+		// It is text/plain, but an attachment.
+		// Content-Type: text/plain; name="test.csv"
+		// Content-Disposition: attachment; filename="test.csv"
+		// Check for attachment only, or inline body is marked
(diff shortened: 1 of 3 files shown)


detect.go
```diff
@@ -63,6 +63,16 @@ func detectTextHeader(header textproto.MIMEHeader, emptyContentTypeIsText bool)
 // detectBinaryBody returns true if the mail header defines a binary body.
 func detectBinaryBody(root *Part) bool {
 	if detectTextHeader(root.Header, true) {
+		// It is text/plain, but an attachment.
+		// Content-Type: text/plain; name="test.csv"
+		// Content-Disposition: attachment; filename="test.csv"
+		// Check for attachment only, or inline body is marked
+		// as attachment, too.
+		mediatype, _, _, _ := ParseMediaType(root.Header.Get(hnContentDisposition))
+		if strings.ToLower(mediatype) == cdAttachment {
```
golangcibot: S1008: should use 'return <expr>' instead of 'if <expr> { return <bool> }; return <bool>' (from `gosimple`)

requaos: Suggestion: `return strings.ToLower(mediatype) == cdAttachment` instead of using if block

jhillyerd:
Thanks!

Issue #131, opened 2019-08-28
Failed to decode BodyPlainText correctly


liangyucheng14:
What I did:
when I try to parse a mail, it failed to parse the bodyPlainText field due to wrong charset.

What I expected:
I expect the body plain text should be 和弟弟
What I got:
But I got garbled, which is caused by that the library use the charset window-1252, but actually should be utf8.

Release or branch I am using:
master

(Please attach a sample message if you feel it will help reproduce the issue)
MIME-Version: 1.0

Content-Type: multipart/alternative; boundary="000000000000b08cae0591253956"

--000000000000b08cae0591253956
Content-Type: text/plain; charset="UTF-8"
Content-Transfer-Encoding: base64

5ZKM5byf5byfDQo=
--000000000000b08cae0591253956
Content-Type: text/html; charset="UTF-8"
Content-Transfer-Encoding: base64

PGRpdiBkaXI9Imx0ciI+5ZKM5byf5byfPC9kaXY+DQo=
--000000000000b08cae0591253956--

jhillyerd:
We use github.com/saintfish/chardet to detect character set and override what is provided in the Content-Type field if it has a high enough confidence; as Content-Type is often wrong.

I have a feeling that message is too short for it to work well.  We might consider adding a minimum threshold number of characters before we invoke chardet.

liangyucheng14:
Agreed. Maybe the plaintext is too short for the select algorithm to give a right encoding type. My suggestion is that we can consider using Content-Type in the mail if its confidence exceed a threshold, which sound more reasonable than only rely on the select algorithm. 

Anyway, thank you for your response!

jhillyerd:
I believe we are already doing that: https://github.com/jhillyerd/enmime/blob/48138bd64d88b19c499fc8dde63d356f66f5b32d/part.go#L21

May need to research how many characters the chardet algorithm needs to be reliable.

requaos:
Per
https://github.com/jhillyerd/enmime/blob/48138bd64d88b19c499fc8dde63d356f66f5b32d/part.go#L283

I don't think we check the content length against that constant

requaos:
We could check the length of the buffer here:
https://github.com/jhillyerd/enmime/blob/48138bd64d88b19c499fc8dde63d356f66f5b32d/part.go#L181
And if it's not long enough, return a new io.Reader from the unmodified
bytes we just read in.

requaos:
[...]

PR #132, opened 2019-08-29
#131 include rune count minimum for chardet

requaos:


requaos:
This is a proposal to resolve #131

jhillyerd:
Looks good to me, merge when ready!

Issue #64, opened 2018-03-18
Feature: Parse headers and stop


jhillyerd:
Inbucket needs to read message headers during delivery, but doesn't need to parse the entire message until somebody tried to view it.

Go's built in header parsing doesn't handle encoded headers.  It would be nice if enmime could just parse the headers and return them.

requaos:
What type of encoding? Do you have a requirement to decode RFC2047 QP and B64 header values to plain-text?

requaos:
I wrote this a while back to preserve the ordinality of the headers with decoded values, however the resultant `Decoded []byte` value is for humans, ie- it's unicode and doesn't respect folding-whitespace or line-length requirements.

```go
type HeadersPreserved struct {
	Decoded []byte
	Headers Headers
}

type Headers []Header

type Header struct {
	Name  string
	Value string
}

func Sort(b []byte) (*HeadersPreserved, error) {
	b = Clean(b)
	tr := textproto.NewReader(bufio.NewReader(bytes.NewReader(b)))
	headers, err := tr.ReadMIMEHeader()
	switch errors.Cause(err) {
	case nil, io.EOF:
	// carry on, io.EOF is expected
	default:
		return nil, err
	}
	bs := bufio.NewScanner(bufio.NewReader(bytes.NewReader(b)))
	res := Headers{}
	bw := &bytes.Buffer{}
	for bs.Scan() {
		line := bs.Text()
		if strings.HasPrefix(line, " ") || strings.HasPrefix(line, "\t") {
			continue
		}
		i := strings.Index(line, ":")
		if i == -1 {
			continue
		}
		header := textproto.CanonicalMIMEHeaderKey(line[:i])
		if len(headers[header]) == 0 {
			// somethings up, we should have already processed all of these, so why are we trying to fetch from an empty list, did we miscount?
			continue
		}
		// pop
		firstValue := headers[header][0]
		// shift
		headers[header] = headers[header][1:]

		h := Header{Name: header, Value: RFC2047parts(firstValue)}
		res = append(res, h)
		bw.WriteString(fmt.Sprintf("%s: %s\r\n", h.Name, h.Value))
	}
	bw.WriteString("\r\n")

	return &HeadersPreserved{
		Decoded: bw.Bytes(),
		Headers: res,
	}, nil
}

func Clean(b []byte) []byte {
	slice := bytes.SplitAfter(b, []byte{'\r', '\n'})
	dest := make([]byte, 0, len(b)+2)
	headers := true
[...]

PR #133, opened 2019-08-30
added public HumanHeadersOnly package-level function

requaos:
#64 Proposed Solution

Obviously the function name is up for debate, but this is the general flow
I noticed that the main lib doesn't yet support recursive rfc2047 encoded header values, so that will need to get incorporated into the ReadEnvelope flow as well

### DIFF
--- a/inspect.go
+++ b/inspect.go
@@ -96,7 +96,7 @@ func rfc2047decode(s string) string {
 	}
 }
 
-// rfc2047recurse is called for if the value contains content encoded in RFC2047 format and decodes it
+// rfc2047recurse is called for if the value contains content encoded in RFC2047 format and decodes it.
 // RFC2047 Example:
 //     `=?UTF-8?B?bmFtZT0iw7DCn8KUwoo=?=`
 func rfc2047recurse(s string) (string, error) {
--- a/inspect.go
+++ b/inspect.go
@@ -12,8 +12,24 @@ import (
 	"github.com/pkg/errors"
 )
 
-func HumanHeadersOnly(b []byte) (map[string]string, error) {
-	b = clean(b)
+var defaultHeadersList = []string{
+	"From",
+	"To",
+	"Sender",
+	"CC",
+	"BCC",
+	"Subject",
+	"Date",
+}
+
+// DecodeHeaders returns a limited selection of mime headers for use by user agents
+// Default header list:
+//   "Date", "Subject", "Sender", "From", "To", "CC" and "BCC"
+//
+// Additional headers provided will be formatted canonically:
+//   h, err := enmime.DecodeHeaders(b, "content-type", "user-agent")
+func DecodeHeaders(b []byte, addtlHeaders ...string) (textproto.MIMEHeader, error) {
+	b = ensureHeaderBoundary(b)
 	tr := textproto.NewReader(bufio.NewReader(bytes.NewReader(b)))
 	headers, err := tr.ReadMIMEHeader()
 	switch errors.Cause(err) {
@@ -22,34 +38,21 @@ func HumanHeadersOnly(b []byte) (map[string]string, error) {
 	default:
 		return nil, err
 	}
-	bs := bufio.NewScanner(bufio.NewReader(bytes.NewReader(b)))
-	res := map[string]string{}
-	for bs.Scan() {
-		line := bs.Text()
-		if strings.HasPrefix(line, " ") || strings.HasPrefix(line, "\t") {
-			continue
+	headerList := append(defaultHeadersList, addtlHeaders...)
+	res := map[string][]string{}
+	for _, header := range headerList {
+		h := textproto.CanonicalMIMEHeaderKey(header)
+		res[h] = make([]string, 0, len(headers[h]))
+		for _, value := range headers[h] {
+			res[h] = append(res[h], rfc2047parts(value))
 		}
-		i := strings.Index(line, ":")
-		if i == -1 {
-			continue
-		}
-		header := textproto.CanonicalMIMEHeaderKey(line[:i])
-		if len(headers[header]) == 0 {
-			// somethings up, we should have already processed all of these, so why are we trying to fetch from an empty list, did we miscount?
-			continue
-		}
-		// pop
-		firstValue := headers[header][0]
-		// shift
-		headers[header] = headers[header][1:]
-
-		res[header] = rfc2047parts(firstValue)
 	}
 
 	return res, nil
 }
 
-func clean(b []byte) []byte {
+// ensureHeaderBoundary scans through an rfc822 document to ensure the boundary between headers and body exists
+func ensureHeaderBoundary(b []byte) []byte {
 	slice := bytes.SplitAfter(b, []byte{'\r', '\n'})
 	dest := make([]byte, 0, len(b)+2)
 	headers := true
--- a/inspect.go
+++ b/inspect.go
@@ -74,7 +74,7 @@ func ensureHeaderBoundary(b []byte) []byte {
 	return dest
 }
 
-// rfc2047decode returns a decoded string if the input uses RFC2047 encoding, otherwise it will return the input
+// rfc2047decode returns a decoded string if the input uses RFC2047 encoding, otherwise it will return the input.
 // RFC2047 Example:
 //     `=?UTF-8?B?bmFtZT0iw7DCn8KUwoo=?=`
 func rfc2047decode(s string) string {
--- a/inspect.go
+++ b/inspect.go
@@ -153,7 +153,7 @@ func fixRFC2047String(s string) string {
 			if eq {
 				inString = true
 			} else {
-				q += 1
+				q++
 			}
 			eq = false
 			sb.WriteRune(v)
--- a/inspect.go
+++ b/inspect.go
@@ -57,7 +57,7 @@ func ensureHeaderBoundary(b []byte) []byte {
 	dest := make([]byte, 0, len(b)+2)
 	headers := true
 	for _, v := range slice {
-		if (bytes.Index(v, []byte{':'}) > -1 || bytes.HasPrefix(v, []byte{' '}) || bytes.HasPrefix(v, []byte{'\t'})) && headers {
+		if headers && (bytes.Index(v, []byte{':'}) > -1 || bytes.HasPrefix(v, []byte{' '}) || bytes.HasPrefix(v, []byte{'\t'})) {
 			dest = append(dest, v...)
 			continue
 		}
--- a/inspect.go
+++ b/inspect.go
@@ -57,7 +57,7 @@ func ensureHeaderBoundary(b []byte) []byte {
 	dest := make([]byte, 0, len(b)+2)
 	headers := true
 	for _, v := range slice {
-		if headers && (bytes.Index(v, []byte{':'}) > -1 || bytes.HasPrefix(v, []byte{' '}) || bytes.HasPrefix(v, []byte{'\t'})) {
+		if headers && (bytes.Contains(v, []byte{':'}) || bytes.HasPrefix(v, []byte{' '}) || bytes.HasPrefix(v, []byte{'\t'})) {
 			dest = append(dest, v...)
 			continue
 		}
--- a/inspect.go
+++ b/inspect.go
@@ -1,0 +1,166 @@
+package enmime
+
+import (
+	"bufio"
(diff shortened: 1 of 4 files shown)


inspect_test.go
```diff
@@ -0,0 +1,64 @@
...
+		}
+	})
+}
+
+func BenchmarkHumanHeadersOnly(b *testing.B) {
+	r := test.OpenTestData("mail", "qp-utf8-header.raw")
+	test, _ := ioutil.ReadAll(r)
+	for i := 0; i < b.N; i++ {
+		h, _ := enmime.HumanHeadersOnly(test)
+		mail.ParseAddressList(h["From"])
```
golangcibot: Error return value of `mail.ParseAddressList` is not checked (from `errcheck`)

inspect_test.go
```diff
@@ -0,0 +1,64 @@
...
+		mail.ParseAddressList(h["To"])
```
golangcibot: Error return value of `mail.ParseAddressList` is not checked (from `errcheck`)

inspect_test.go
```diff
@@ -0,0 +1,64 @@
...
+		_ = h["Subject"]
+	}
+}
+
+func BenchmarkReadEnvelope(b *testing.B) {
+	r := test.OpenTestData("mail", "qp-utf8-header.raw")
+	for i := 0; i < b.N; i++ {
+		env, _ := enmime.ReadEnvelope(r)
+		env.AddressList("From")
```
golangcibot: Error return value of `env.AddressList` is not checked (from `errcheck`)

inspect_test.go
```diff
@@ -0,0 +1,64 @@
...
+		env.AddressList("To")
```
golangcibot: Error return value of `env.AddressList` is not checked (from `errcheck`)

inspect.go
```diff
@@ -0,0 +1,166 @@
...
+
+	return res, nil
+}
+
+func clean(b []byte) []byte {
+	slice := bytes.SplitAfter(b, []byte{'\r', '\n'})
+	dest := make([]byte, 0, len(b)+2)
+	headers := true
+	for _, v := range slice {
+		if (bytes.Index(v, []byte{':'}) > -1 || bytes.HasPrefix(v, []byte{' '}) || bytes.HasPrefix(v, []byte{'\t'})) && headers {
```
golangcibot: S1003: should use bytes.Contains(v, []byte{':'}) instead (from `gosimple`)

requaos:
Preliminary benchmarks results put this ~%40 faster than calling ReadEnvelope for just those three headers. In this function we still read the whole `[]byte` when it can just halt and return when the header/body boundary is reached

inspect.go
```diff
@@ -0,0 +1,166 @@
...
+	"io"
+	"mime"
+	"net/textproto"
+	"strings"
+
+	"github.com/jhillyerd/enmime/internal/coding"
+	"github.com/pkg/errors"
+)
+
+func HumanHeadersOnly(b []byte) (map[string]string, error) {
```
jhillyerd: It doesn't look like we are actually limiting the returned headers to a human list, and we probably don't need to do that anyway.  Should we just call it `DecodeHeaders` instead?

inspect.go
```diff
@@ -0,0 +1,166 @@
...
+	b = clean(b)
+	tr := textproto.NewReader(bufio.NewReader(bytes.NewReader(b)))
+	headers, err := tr.ReadMIMEHeader()
+	switch errors.Cause(err) {
+	case nil, io.EOF:
+	// carry on, io.EOF is expected
+	default:
+		return nil, err
+	}
+	bs := bufio.NewScanner(bufio.NewReader(bytes.NewReader(b)))
```
jhillyerd: Scanner has it's own buffer, don't need the extra bufio.Reader.

inspect.go
```diff
@@ -0,0 +1,166 @@
...
+		if strings.HasPrefix(line, " ") || strings.HasPrefix(line, "\t") {
+			continue
+		}
+		i := strings.Index(line, ":")
+		if i == -1 {
+			continue
+		}
+		header := textproto.CanonicalMIMEHeaderKey(line[:i])
+		if len(headers[header]) == 0 {
+			// somethings up, we should have already processed all of these, so why are we trying to fetch from an empty list, did we miscount?
```
jhillyerd: Comment is a little confusing, maybe mention it's populated by ReadMIMEHeader?

inspect.go
```diff
@@ -0,0 +1,166 @@
...
+			continue
+		}
+		// pop
```
jhillyerd: I think it would be more useful to explain why we are doing this (preserving header order?) rather than just pop/shift.

inspect.go
```diff
@@ -0,0 +1,166 @@
...
+		// shift
+		headers[header] = headers[header][1:]
+
+		res[header] = rfc2047parts(firstValue)
+	}
+
+	return res, nil
+}
+
+func clean(b []byte) []byte {
```
jhillyerd: Add descriptive comment and rename to something more specific

inspect.go
jhillyerd: Move headers to the front of the if clause to save indexing into byte slice unnecessarily.  Or we move this if block into the existing `if headers` block below to avoid testing it twice; if it doesn't make things unreadable.

inspect.go
```diff
@@ -0,0 +1,166 @@
...
+				continue
+			}
+		}
+		dest = append(dest, v...)
+	}
+
+	return dest
+}
+
+// rfc2047parts checks if the value contains content encoded in RFC2047 format
```
jhillyerd: I don't think "checks" is the correct term if the func returns a string, I'd expect that to be returning a bool.

inspect.go
```diff
@@ -0,0 +1,166 @@
...
+	dec := new(mime.WordDecoder)
+	dec.CharsetReader = coding.NewCharsetReader
+	header, err := dec.DecodeHeader(input)
+	if err != nil {
+		return input, err
+	}
+	return header, nil
+}
+
+func fixRFC2047String(s string) string {
```
jhillyerd: Needs comment on what type of fix it is making.

inspect_test.go
```diff
@@ -0,0 +1,94 @@
...
+		if err != nil {
+			b.Fatal(err)
+		}
+		_, err = env.AddressList("To")
+		if err != nil {
+			b.Fatal(err)
+		}
+		env.GetHeader("Subject")
+		// reset reader for next run
+		_, err = reusedReader.Seek(0, io.SeekStart)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

inspect.go
requaos: Now limited to a human list to reduce execution time, but still renamed to `DecodeHeaders` and added variadic second parameter for supplying additional header keys the consumer wants in the returned map.

inspect.go
requaos: Not preserving order for this use case, so scanner isn't required anymore

inspect.go
requaos: not preserving order anymore

inspect.go
```diff
@@ -0,0 +1,169 @@
...
+	return res, nil
+}
+
+// ensureHeaderBoundary scans through an rfc822 document to ensure the boundary between headers and body exists
+func ensureHeaderBoundary(b []byte) []byte {
+	slice := bytes.SplitAfter(b, []byte{'\r', '\n'})
+	dest := make([]byte, 0, len(b)+2)
+	headers := true
+	for _, v := range slice {
+		if headers && (bytes.Index(v, []byte{':'}) > -1 || bytes.HasPrefix(v, []byte{' '}) || bytes.HasPrefix(v, []byte{'\t'})) {
```
golangcibot: S1003: should use bytes.Contains(v, []byte{':'}) instead (from `gosimple`)

inspect.go
```diff
@@ -0,0 +1,171 @@
...
```
golangcibot: S1003: should use bytes.Contains(v, []byte{':'}) instead (from `gosimple`)

inspect.go
requaos: this was removed

requaos:
Much faster now
```
BenchmarkHumanHeadersOnly-4   	  100000	     19347 ns/op
BenchmarkReadEnvelope-4       	    1000	   1213698 ns/op

File: inspect.go (hunk 6 above)
Comment (requaos) [reply inspect.go]: Since function signature has changed, I will resolve this conversation and look forward to new proposed changes on the latest commit.

File: inspect.go [hunk 16]
```diff
@@ -0,0 +1,171 @@
...
+	b = ensureHeaderBoundary(b)
+	tr := textproto.NewReader(bufio.NewReader(bytes.NewReader(b)))
+	headers, err := tr.ReadMIMEHeader()
+	switch errors.Cause(err) {
+	case nil, io.EOF:
+	// carry on, io.EOF is expected
+	default:
+		return nil, err
+	}
+	headerList := append(defaultHeadersList, addtlHeaders...)
```
jhillyerd: This append risks modifying `defaultHeadersList`, should copy it first.  Probably not thread-safe either.

inspect.go
```diff
@@ -0,0 +1,171 @@
...
+				continue
+			}
+		}
+		dest = append(dest, v...)
+	}
+
+	return dest
+}
+
+// rfc2047decode returns a decoded string if the input uses RFC2047 encoding, otherwise it will return the input
```
jhillyerd:
```suggestion
// rfc2047decode returns a decoded string if the input uses RFC2047 encoding, otherwise it will return the input.
```

inspect.go
```diff
@@ -0,0 +1,171 @@
...
+		switch err {
+		case nil:
+			continue
+		default:
+			return s
+		}
+	}
+}
+
+// rfc2047recurse is called for if the value contains content encoded in RFC2047 format and decodes it
```
jhillyerd:
```suggestion
// rfc2047recurse is called for if the value contains content encoded in RFC2047 format and decodes it.
```

inspect.go
```diff
@@ -0,0 +1,171 @@
...
+		return input, err
+	}
+	return header, nil
+}
+
+// fixRFC2047String removes the following characters from charset and encoding segments of an RFC2047 string:
+//  '\n', '\r' and ' '
+func fixRFC2047String(s string) string {
+	inString := false
+	eq := false
```
jhillyerd: Comment indicating that eq tracks when we've seen the initial = in a string (... I think?)

inspect.go
```diff
@@ -0,0 +1,171 @@
...
+	q := 0
```
jhillyerd: Comment indicating q tracks the number of question marks we've seen

inspect.go
requaos: I just gave the variable a long descriptive name: `isWithinTerminatingEqualSigns`

inspect.go
requaos: I just gave the variable a long descriptive name: `questionMarkCount`

inspect.go
requaos: done

requaos:
Once this is merged, we will be able to incorporate the recursive rfc2047 into normal flow and resolve #134 with a refactor of `rfc2047AttributeName` like this:
```
func rfc2047AttributeName(name string) string {
	if !strings.Contains(s, "=?") && !strings.Contains(s, "?=") {
		// the RFC2047 chunk is either malformed or is not an RFC2047 chunk
		return name
	}
	
	// Copy the string so we can return the original if we encounter any issues.
	s := name

	result := &strings.Builder{}

	_, err := result.WriteString(rfc2047decode(s))
	if err != nil {
		return name
	}

	keyValuePair := strings.SplitAfter(result.String(), "=")
	if len(keyValuePair) < 2 {
		return result.String()
	}
	// Add quotes as needed
	if !strings.HasPrefix(keyValuePair[1], "\"") {
		keyValuePair[1] = fmt.Sprintf("\"%s", keyValuePair[1])
	}
	if !strings.HasSuffix(keyValuePair[1], "\"") {
		keyValuePair[1] = fmt.Sprintf("%s\"", keyValuePair[1])
	}
	return strings.Join(keyValuePair, "")
}

requaos:
@jhillyerd If everything looks good here, feel free to merge

jhillyerd:
Thanks!

PR #135, opened 2019-09-06
WIP: Line wrapping for text/plain content

### LINKED ISSUE #63: Wrap plain text and HTML parts
enmime does not enforce a line-length on text parts.  It should wrap the content in a way that preserves the original bytes upon decoding.

requaos:
Implementing text wrapping for `text/html` is a bit more involved, such that we don't break mid-tag. We should create an issue to track wrapping html content for the future.

closes #63


encode.go
```diff
@@ -213,3 +216,22 @@ func setParamValue(p map[string]string, k, v string) {
 		p[k] = v
 	}
 }
+
+func wrapContent(text []byte, b *bufio.Writer) error {
+	lineLen := lineWrapLength
+	beginning := true
+	for len(text) > 0 {
+		if !beginning {
+			b.Write(crnl)
```
golangcibot: Error return value of `b.Write` is not checked (from `errcheck`)

requaos:
At present this just uses the same wrapping implementation as the base64 content. @jhillyerd from my reading this won't matter, but let em know if you would prefer to see this not wrap mid-word.

Still work to do writing more tests for increased coverage instead of decreasing...

jhillyerd:
Thinking about this a bit...  the only reason to force a wrap before 76-80 characters is to improve the readability of the text in a plain text email program, or when viewing the source - if we split the line in the middle of the word, we aren't improving the readability.

It may be better to pick an wider (but <= 998) number so that if the email is wrapped at something slightly higher than 76, we allow it to pass through, otherwise we hard wrap to remain compliant with the spec.

There is some interesting reading in [rfc3676 4.2](https://tools.ietf.org/html/rfc3676#section-4.2), but I don't know much about format=flowed.

requaos:
I will follow the technique described in paragraph six of rfc3676 4.2, the "newer" one. I feel there is enough feature description in the following sections to implement format-(fixed/flowed), delsp-(yes/no), space-stuffing, quoting and even take the Usenet signature separator into consideration.

requaos:
Looks like there is a reasonable implementation, for decoding, flowed text in k9mail. Porting that shouldn't take too long, but creating an encoder for it will take longer.

requaos:
Looks like html is out of scope for this:

`Format=flowed applies solely to plain-text messages. HTML messages already have something functionally equivalent to f=f: the <BLOCKQUOTE> element, which does what it says: It quotes blocks of text. When f=f mailers that also can handle HTML encounter <BLOCKQUOTE> text, it’s usually marked up with the same excerpt bars we’re familiar with from f=f. Format=flowed isn’t actually at work there, but since <BLOCKQUOTE> text flows nicely when you resize a window, the effect is the same.`

requaos:
@jhillyerd I ported the feature verbatim and wrote one test around the only example in the rfc. I need to find a preferred implementation, like either pine or mutt, where I can generate some testdata samples programmatically.

Issue #134, opened 2019-09-06
PANIC: github.com/jhillyerd/enmime@v0.6.0/header.go:637


beiping96:
```
runtime error: slice bounds out of range 1024 goroutine 5343743 [running]:
panic(0xb55c00, 0x142aff0)
        /usr/local/go/src/runtime/panic.go:522 +0x1b5
github.com/jhillyerd/enmime.rfc2047AttributeName(0xc009828319, 0xa9, 0xc2c6fe, 0x1)
        /home/go/pkg/mod/github.com/jhillyerd/enmime@v0.6.0/header.go:637 +0x94e
github.com/jhillyerd/enmime.fixMangledMediaType(0xc009828300, 0x168, 0xc2c6fc, 0x1, 0x0, 0xd51f40)
        /home/go/pkg/mod/github.com/jhillyerd/enmime@v0.6.0/header.go:298 +0x49c
github.com/jhillyerd/enmime.ParseMediaType(0xc009828300, 0x168, 0xc, 0xc009828300, 0x168, 0x1000, 0xc009683000, 0x0, 0x0, 0xc0003a2700)
        /home/go/pkg/mod/github.com/jhillyerd/enmime@v0.6.0/header.go:231 +0x3cf
github.com/jhillyerd/enmime.(*Part).setupHeaders(0xc0096b6000, 0xc010f66780, 0x0, 0x0, 0x0, 0x0)
        /home/go/pkg/mod/github.com/jhillyerd/enmime@v0.6.0/part.go:116 +0xdc
github.com/jhillyerd/enmime.parseParts(0xc003840200, 0xc01887e7e0, 0xa, 0
```

requaos:
@beiping96 Could you please provide some data to reproduce this?

requaos:
Here is a playground for reproducing this issue: https://play.golang.org/p/fjDiuq5qwVP

beiping96:
https://play.golang.org/p/ghLy5zbMVzC

Here you are

requaos:
@beiping96 have you tested against `develop` branch? This should be resolved by #133

pedrinimm:
@requaos The bug is on the develop branch, method rfc2047AttributeName the error is on referencing indexed on the slice that do not exist.

If you want a string that triggers it here: ?=p?=?

If you isolate the function and pass that as string you have the error.

requaos:
[...]

PR #136, opened 2019-10-15
use rfc2047decode in fixMangledMediaType for #134

requaos:


requaos:
@jhillyerd Not sure how to reproduce the error that travis is failing on, I have run `go test -race ./...` with a `1.12` and a `1.13` GOROOT, please advise

requaos:
https://github.com/golang/go/issues/34540
net/http (through net/textproto) used to accept and normalize invalid HTTP/1.1 headers with a space before the colon, in violation of RFC 7230. If a Go server is used behind a reverse proxy that accepts and forwards but doesn't normalize such invalid headers, the reverse proxy and the server can interpret the headers differently. This can lead to filter bypasses or request smuggling, the latter if requests from separate clients are multiplexed onto the same connection by the proxy. Such invalid headers are now rejected by Go servers, and passed without normalization to Go client applications.

This issue is CVE-2019-16276 and is fixed in Go 1.13.1 and Go 1.12.10.

dcormier:
Might want to check if `\t` is handled or not. I thought when I glanced at the golang change that that case is also different now.

PR #139, opened 2019-12-21
mime: invalid media parameter

requaos:
While we handle media parameters with missing values if they are the last parameter, this fix allows for the handling of media parameters missing value in the middle of the parameter list.

### DIFF
--- a/header.go
+++ b/header.go
@@ -383,6 +383,13 @@ findValueStart:
 
 			break findValueStart
 
+		case ';':
+			if value.Len() == 0 {
+				value.WriteString(`"";`)
+			}
+
+			break findValueStart
+
 		default:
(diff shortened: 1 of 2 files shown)


requaos:
Each of the two above commits, solve a different problem. Both problems result in the same error.

1. `text/html; charset=; format=flowed` <-- the middle parameter was not getting fixed
2. `multipart/mixed; boundary=?BE023287982-wefweiuh38839` <-- The first character '?' was not being included in the check to see if the parameter value needs to be quoted.

jhillyerd:
Thanks.

BTW, I think your test coverage increase may have drawn some recent attention: https://starcharts.herokuapp.com/jhillyerd/enmime

PR #141, opened 2019-12-23
Coverage and cleanup

requaos:


requaos:
@jhillyerd This also cleans up a fabricated error case that was designed to catch malformed header boundaries with the body in conjunction with an `UnreadByte()` error of `io.UnexpectedEOF`. Since Go 1.13 release, when we refactored this code, the UnreadByte error is gone and so we no longer need this custom error exchange between the boundaryreader and the other code. Plus, removing unreachable code increases coverage 👍

inspect.go
```diff
@@ -123,34 +121,16 @@ func rfc2047recurse(s string) (string, error) {
...
-	if err != nil {
-		return val, err
-	}
-	if val == s {
-		val, err = decodeHeaderWithError(fixRFC2047String(val))
-		if err != nil {
-			return val, err
-		}
-		if val == s {
+	val := s
```
golangcibot: ineffectual assignment to `val` (from `ineffassign`)

requaos:
@jhillyerd I guess this turned out to be a solid cleanup sweep. Getting that last 4+ percent is going to be quite challenging, so maybe sometime next year we can get around to it.

jhillyerd:
Lot of nice simplifications in here, hopefully it makes the code easier to follow.

PR #144, opened 2020-01-03
Handle very empty MIME part

dcormier:
When one MIME boundary immediately follows another, treat it as an empty part.

This comes from an issue I saw in a production environment where there was an email where a MIME part delimiter was immediately followed by a terminator. Like this:

```txt
Content-type: multipart/mixed; boundary="the-boundary"

--the-boundary
Content-type: text/plain

Some content.

--the-boundary
--the-boundary--
```

### DIFF
--- a/boundary.go
+++ b/boundary.go
@@ -21,6 +21,7 @@ var errNoBoundaryTerminator = stderrors.New("expected boundary not present")
 type boundaryReader struct {
 	finished         bool          // No parts remain when finished
 	partsRead        int           // Number of parts read thus far
+	atPartStart      bool          // Whether the current part is at its beginning
 	r                *bufio.Reader // Source reader
 	nlPrefix         []byte        // NL + MIME boundary prefix
 	prefix           []byte        // MIME boundary prefix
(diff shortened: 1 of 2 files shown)


requaos:
```
		if b.atPartStart && n > 0 {
			b.atPartStart = false
		}
```
can be simplified to 
```
		b.atPartStart = b.atPartStart && n == 0
```

dcormier:
Why not only set the variable only if it needs to be set?

requaos:
The block doesn't ever evaluate to true at `boundary.go:77`, which makes `boundary.go:78` unreachable.
For this one block, if we change it to an assignment statement instead of an if-block, we get to keep the business logic without giving up the ~0.06% in code-coverage or readability.

dcormier:
Yes, that would increase the covered code in the coverage report, but it would also hide the fact that there's a condition not exercised by the current test suite.

From what I've been able to tell, I don't _think_ there's a condition where it could get into that block. But I'm not 100% sure (and I have tried to write a test for it, just to be sure), and maybe something will change in the future that would cause that block to be exercised.

If you still want me to make the change to hide the untested condition, I'll go ahead and do that. I just wanted to make sure it was clear.

requaos:
@dcormier Those are very good points. I have adjusted one of the boundaryReader tests to hit that block of code instead and opened a PR to your branch: [Test Cover](https://github.com/dcormier/enmime/pull/1)

requaos:
This looks ready to merge

jhillyerd:
Thanks folks!

PR #145, opened 2020-01-03
Read into destination by length, not capacity

dcormier:
Fixes the `boundaryReader` implementation of `io.Reader` to fill the slice based on its length, not its capacity. As per [the `io.Reader` documentation](https://pkg.go.dev/io?tab=doc#Reader).


jhillyerd:
Good catch, thank you!

PR #148, opened 2020-02-21
Adds Part.ContentTypeParams to allow custom params

pzeinlinger:
Custom ContentTypeParams are required in certain situations.
 E.g. RFC3156 (MIME Security for OpenPGP) requires
 protocol="application/pgp-encrypted" as Content-Type param.

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -100,6 +100,9 @@ func (p *Part) setupMIMEHeaders() transferEncoding {
 	if p.ContentType != "" {
 		// Build content type header.
 		param := make(map[string]string)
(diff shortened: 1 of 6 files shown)


jhillyerd:
Thanks, this is a good idea!

Can you add a unit test?  Perhaps extend https://github.com/jhillyerd/enmime/blob/master/encode_test.go#L34 and its golden file.  That way we can ensure any future refactoring doesn't break this functionality.

pzeinlinger:
Sure thing!

jhillyerd:
Looks good, thank you!

PR #149, opened 2020-03-05
Header line incorrectly beginning with a space

requaos:
results in "mime: invalid media parameter" error

This is a best effort attempt to correct an invalid continuation.


jhillyerd:
Looks good, thanks.

Issue #153, opened 2020-05-21
Trailing comma for Content-Type value causes an error


joekamibeppu:
Hi James, thanks for providing and maintaining this great tool! I ran into this while using it and wanted to see what you thought.

What I did: 

Run `enmime.ReadEnvelope()` on a raw email containing the header `Content-Type: text/plain;`. Note the trailing `;` at the end, with nothing following it.

What I expected:

The raw email to be parsed successfully.

What I got:

An error saying, `mime: invalid media parameter`

Release or branch I am using:

Release `v0.8.0`

If this is indeed an issue, I believe I have a small fix ready. I could open a PR against the `develop` branch.

Thanks!

dcormier:
> If this is indeed an issue, I believe I have a small fix ready. I could open a PR against the develop branch.

[Contributions are encouraged.](https://github.com/jhillyerd/enmime/blob/develop/CONTRIBUTING.md)

jhillyerd:
Thanks for reporting Joe!  We'd definitely appreciate a fix, especially if it includes a unit test so we can prevent a regression in the future.

PR #154, opened 2020-05-22
Handle trailing separators in header

joekamibeppu:
Hello. This is a fix for https://github.com/jhillyerd/enmime/issues/153.

If my understanding is correct, here is what was happening. When the `Content-Type` header value was something like `text/plain;`, it would turn into `text/plain;=not-a-param-value`, then `=not-a-param-value` parsed on its own would cause a `mime: invalid media parameter` error in turn.

Thanks!

### DIFF
--- a/header.go
+++ b/header.go
@@ -287,6 +287,11 @@ func fixMangledMediaType(mtype, sep string) string {
 				}
 			}
 		default:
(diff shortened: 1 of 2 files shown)


jhillyerd:
Thanks!

PR #157, opened 2020-06-24
fix: using dfs locate html body

huaconghub:
i found outlook mail service genrate an eml struct like:
- multipart/mixed
- - multipart/related
- - - multipart/alternative
- - - - text/plain (real mail content)
- - - - text/html (real mail content)
- - multipart/alternative
- - - text/plain (contact info)
- - - text/html (contact info)

bfs(first) will locate the wrong html(contact info), i checked Apache mail common, using dfs(first) to locate html, btw macOS mail client and outlook client dfs all content, not just first

### DIFF
--- a/envelope.go
+++ b/envelope.go
@@ -300,7 +300,7 @@ func parseMultiPartBody(root *Part, e *Envelope) error {
 	}
 
 	// Locate HTML body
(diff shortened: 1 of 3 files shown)


jhillyerd:
Makes sense to me, thank you!

Issue #152, opened 2020-04-28
Error decoding attachment file names in utf-8


int01:
Hi,

Any file name in the attachment is decoded incorrectly if it has at least one character with a length of more than one byte.

The fixUnescapedQuotes function in file header.go is not working properly.

This code
```
		param = param[startingQuote+1 : closingQuote]
		escaped := false
		for strIdx := range param {
			switch param[strIdx] {
			case '"':
				// We are inside of a quoted string, so lets escape this guy if it isn't already escaped.
				if !escaped {
					sb.WriteByte('\\')
					escaped = false
				}
				sb.WriteByte(param[strIdx])
			case '\\':
				// Something is getting escaped, a quote is the only char that needs
				// this, so lets assume the following char is a double-quote.
				escaped = true
				sb.WriteByte('\\')
			default:
				escaped = false
				sb.WriteByte(param[strIdx])
			}
		}

```
should be replaced with this one (WriteByte -> WriteRune).
```
		param = param[startingQuote+1 : closingQuote]
		escaped := false
		for _, c := range param {
			switch c {
			case '"':
				// We are inside of a quoted string, so lets escape this guy if it isn't already escaped.
				if !escaped {
					sb.WriteByte('\\')
					escaped = false
				}
				sb.WriteRune(c)
			case '\\':
				// Something is getting escaped, a quote is the only char that needs
				// this, so lets assume the following char is a double-quote.
				escaped = true
				sb.WriteByte('\\')
			default:
				escaped = false
				sb.WriteRune(c)
			}
		}

```

And one more value for header_test.go:
```
		{
			input: "application/rtf; charset=iso-8859-1; name=\"тест.rtf;\".rtf",
			want:  "application/rtf; charset=iso-8859-1; name=\"\\\"тест.rtf;\\\".rtf\"",
		},

```
Thanks.

jhillyerd:
Please submit a pull request against the develop branch.

requaos:
@int01 the document is `us-ascii` at this point.
This should be invalid: `Content-Type: application/rtf; charset=iso-8859-1; name="тест.rtf;".rtf`
RFC2047 makes provisions for this via quoted-printable or base64 encoding. An example of the base64 variant would be encoded like this: `Content-Type: application/rtf; charset=iso-8859-1; name="=?UTF-8?B?0YLQtdGB0YI=?=.rtf;".rtf`

requaos:
[...]

PR #165, opened 2020-08-25
fixUnescapedQuotes fixed for higher Unicode codepoints

pavelbazika:
String is expected to be iterated by bytes, but was by runes.


jhillyerd:
Looks good to me, thanks!  Is this a fix for #152?

pavelbazika:
Sorry for the long delay, I think it fixes #152 .

Issue #159, opened 2020-07-14
Formatted header breaks email readability when received on mail user agent


pacellig:
> Release or branch I am using: v0.8.1 

Parsing formatted headers using the library leads to an email that is no longer correctly parsable by mail user agents. A sample header is attached [here](https://github.com/jhillyerd/enmime/files/4916869/sample_header.txt) for readability with the original formatting.

This has been introduced in [this](https://github.com/jhillyerd/enmime/pull/149/commits/eebd72dcaea5944152993d68671e93953893d1c1) commit, function `readHeader` in header.go file, lines 107-113.

What is obtained after parsing the incoming email:
`
0.00 BSF_BESS_OUTBOUND      META: BESS Outbound
0.00 HTML_MESSAGE           BODY: HTML included in message
0.01 SUBJ_ALL_CAPS          META: Subject is all capitals
0.20 BSF_SC0_SAXXX          META: Custom Rule BSF_SC0_SAXXX
X-Bess-Outbound-Spam-Report: Code version 3.2, rules version 1.2.3.4 [from scanhost.whatever.wherever.com] Rule breakdown below pts rule name              description ---- ---------------------- --------------------------------`

Which breaks parsing and displaying.

pacellig:
Further elaboration on this issue: the test case implemented in handler_test.go#513 is wrong, as it does not respect the long header fields definition of [RFC2822](https://tools.ietf.org/html/rfc2822#section-2.2.3). 

`		{
			input:   "X-Not-Continuation: line1=foo;\n X-Next-Header: bar\n",
			hname:   "X-Not-Continuation",
			want:    "line1=foo;",
			correct: true,
		},
`

The correct result in this case would be:
`		{
			input:   "X-Not-Continuation: line1=foo;\n X-Next-Header: bar\n",
			hname:   "X-Not-Continuation",
			want:    "line1=foo;\n X-Next-Header: bar\n",
			correct: true,
		},
`

jhillyerd:
I think we need to do better than just reverting #149 - as it appears it is preventing a full-stop error, which is worse than a header with misformatted whitespace in my opinion.

jhillyerd:
cc @requaos in case he has ideas.

Looking at the code in #149 it cares about where the colon is in relation to equal signs, but I think it will accept a header with spaces in its name, instead of considering that a continuation.  That seems problematic, as I'd expect it to error out later.

pacellig:
[...]

PR #166, opened 2020-09-10
Feature/159

pacellig:
Fix reported issue #159 and test cases

### DIFF
--- a/header.go
+++ b/header.go
@@ -107,8 +107,7 @@ func readHeader(r *bufio.Reader, p *Part) (textproto.MIMEHeader, error) {
 			sTrimmed := textproto.TrimBytes(s)
 			firstTrimmedColon := bytes.IndexByte(sTrimmed, ':')
 			firstTrimmedSpace := bytes.IndexAny(sTrimmed, " \t\n\r")
-			if firstTrimmedSpace > 0 && firstTrimmedColon > 0 &&
-				firstTrimmedSpace == firstTrimmedColon+1 {
+			if firstTrimmedColon > 0 && firstTrimmedSpace == firstTrimmedColon+1 {
 				s = sTrimmed
 				firstColon = firstTrimmedColon
 			} else {
--- a/header.go
+++ b/header.go
@@ -101,21 +101,10 @@ func readHeader(r *bufio.Reader, p *Part) (textproto.MIMEHeader, error) {
 		firstColon := bytes.IndexByte(s, ':')
 		firstSpace := bytes.IndexAny(s, " \t\n\r")
 		if firstSpace == 0 {
-			// If there is a single colon followed immediately by a space and there
-			// is no equal sign, or comes before any equal sign, then it should not
-			// be considered as a continuation.
-			colonSpaceIdx := bytes.Index(s, []byte{':', ' '})
-			equalIdx := bytes.IndexByte(s, '=')
-			if bytes.Count(s, []byte{':'}) == 1 && colonSpaceIdx != -1 &&
-				(equalIdx == -1 || equalIdx > colonSpaceIdx) {
-				s = textproto.TrimBytes(s)
-				firstColon = bytes.IndexByte(s, ':')
-			} else {
-				// Starts with space: continuation
-				buf.WriteByte(' ')
-				buf.Write(textproto.TrimBytes(s))
-				continue
-			}
+			// Starts with space: continuation
+			buf.WriteByte(' ')
+			buf.Write(textproto.TrimBytes(s))
+			continue
 		}
 		if firstColon == 0 {
 			// Can't parse line starting with colon: skip
--- a/header.go
+++ b/header.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"mime"
 	"net/textproto"
-	"regexp"
 	"strings"
 
 	"github.com/jhillyerd/enmime/internal/coding"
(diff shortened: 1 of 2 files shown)


dcormier:
@requaos, how's your memory on this one? I'm sure this change is there because of something you encountered.

header.go
```diff
@@ -91,6 +92,11 @@ func readHeader(r *bufio.Reader, p *Part) (textproto.MIMEHeader, error) {
 	buf := &bytes.Buffer{}
 	tp := textproto.NewReader(r)
 	firstHeader := true
+	headerDeclarationRegex, err := regexp.Compile("^[a-zA-Z0-9\\-]+\\: ")
```
pacellig, line 95: This could also be a global variable `var headerDeclarationRegex = regex.MustCompile("^[a-zA-Z0-9\\-]+\\: ")`, saving some computation time. Drawback, should it ever fail it will cause a panic rather than an error that can be handled time by time

header.go
```diff
@@ -101,19 +107,17 @@ func readHeader(r *bufio.Reader, p *Part) (textproto.MIMEHeader, error) {
...
-			equalIdx := bytes.IndexByte(s, '=')
-			if bytes.Count(s, []byte{':'}) == 1 && colonSpaceIdx != -1 &&
-				(equalIdx == -1 || equalIdx > colonSpaceIdx) {
-				s = textproto.TrimBytes(s)
+			// If the line begins with some space, followed by a header-like
+			// string (any combination of upper and lower case letters,
+			// numbers and dash sign), then it should not be considered
+			// as a continuation but as a new header.
+			sTrimmed := textproto.TrimBytes(s)
+			if firstSpace < firstColon && headerDeclarationRegex.Match(sTrimmed) {
```
dcormier, line 115: This is actually quite a bit more restrictive than the logic that was there before, which didn't restrict header names to `a-zA-Z0-9-`. I think this should continue to be as lenient.

Something more like:
```suggestion
			// Get the index of the first whitespace (if any) after the apparent folding whitespace
			firstSpace := bytes.IndexAny(sTrimmed, " \t\n\r")
			if firstSpace > 0 && firstSpace < firstColon {
```

I haven't tested that, but that's the jist of the idea.

jhillyerd: I agree with Daniel, would prefer to avoid the regex.

pacellig: I partially agree, I am ok in removing the regex and doing some 'magic' in relative positions between space and colon, but would still keep the constraint on the first space right after the colon. This is to avoid problems with i.e. DKIM signatures, but I will update the PR so that it's clear what I mean.

dcormier: > keep the constraint on the first space right after the colon.

I'm not sure that's a good idea. I expect headers without a space after the `:` are often encountered.

I also checked [the RFC](https://tools.ietf.org/html/rfc5322) to see what it said about whether that space is even required, and couldn't find a mention of it (someone please prove me wrong).

But to make it even worse, I found [this gem](https://tools.ietf.org/html/rfc5322#section-4.5) (regarding headers):

> Also, any amount of white space is allowed before the ":" at the end of the field name.

So I guess the logic needs to allow non-folding white space as long as it's right before the ':' of the header.

Ugh.

pacellig: Don't get me wrong, this is not per RFC of course, no space is required in a header declaration. 
But as well what we are trying to achieve here is not RFC compliant, so this is being another best-effort attempt to parse some headers as new declarations whereas, according to RFC, should be considered as folded headers.

> Also, any amount of white space is allowed before the ":" at the end of the field name.
This is a good one, and of course would fail with the code here. Nice.

To recap a bit, since this is already an out of standard parsing attempt, I think the less damage solution should be implemented, that means (my personal opinion about this):
 since these headers should not be treated as new at all, if the library manages to cover the most commons header declarations without breaking the parsing of standard headers, it would be a win.

I think the code which is currently on master is worse than the proposed solutions, because it's looking for an '=' sign, attempting to fix only some very specific cases. But again, this is just my personal point of view :)

jhillyerd: I agree the equals sign detection is probably not what we want.  I don't have much time to think or review until the weekend.

pacellig: Any feedback about this? Thanks! :)

header.go
```diff
@@ -101,19 +101,20 @@ func readHeader(r *bufio.Reader, p *Part) (textproto.MIMEHeader, error) {
...
-				(equalIdx == -1 || equalIdx > colonSpaceIdx) {
-				s = textproto.TrimBytes(s)
-				firstColon = bytes.IndexByte(s, ':')
+			// Do not consider as a continuation if the first whitespace
+			//  in the trimmed line comes right after a colon which,
+			//  in turn, follows at least one character.
+			sTrimmed := textproto.TrimBytes(s)
+			firstTrimmedColon := bytes.IndexByte(sTrimmed, ':')
+			firstTrimmedSpace := bytes.IndexAny(sTrimmed, " \t\n\r")
+			if firstTrimmedSpace > 0 && firstTrimmedColon > 0 &&
```
jhillyerd, line 110: `firstTrimmedSpace > 0` -- I think you can eliminate this check, as the == +1 below implies it.

jhillyerd (changes requested):

jhillyerd:
Thanks for all your effort on this one!

pacellig:
Thank you for the valuable feedbacks!

pedrinimm:
@jhillyerd Do you think there could be a v0.8.2 soon given this contribution to the package?
Thanks

jhillyerd:
Yeah, we were overdue.  done!

PR #140, opened 2019-12-21
mime: expected slash after first token

requaos:
Found occurrences of a media parameter list without a mime type, this will inject a safe default mime type.

PR #170, opened 2020-10-26
Improve performance of boundaryReader

### LINKED ISSUE #169: Huge number of allocations in boundaryReader.Read
Hi! I found a huge perfomance issue. In the bondaryReader.Read loop, a single byte is copied from one reader to another using CopyN. Internally this allocates a new LimitedReader on every call.
https://github.com/jhillyerd/enmime/blob/2a0b399b27a8c8a689af62304229c0223ad2c9b0/boundary.go#L149
This can be easily seen using pprof. I replaced CopyN with Read/WriteByte methods and this reduced the cpu usage of our production service by a factor of three.

dcormier:
Explicitly read/write a single byte rather than using `io.CopyN` to copy a single byte.

Before the change:

```
BenchmarkBoundaryReader-12    	  544516	      1909 ns/op	    6704 B/op	      15 allocs/op
```

After:
```
BenchmarkBoundaryReader-12    	 1039579	      1067 ns/op	    4495 B/op	       7 allocs/op
```

From 15 allocs/op down to 7.

Fixes #169.

### DIFF
--- a/boundary.go
+++ b/boundary.go
@@ -146,13 +146,19 @@ func (b *boundaryReader) Read(dest []byte) (n int, err error) {
 			}
 		}
 
-		_, err = io.CopyN(b.buffer, b.r, 1)
+		next, err := b.r.ReadByte()
 		if err != nil {
 			// EOF is not fatal, it just means that we have drained the reader.
-			if errors.Cause(err) == io.EOF {
+			if errors.Is(err, io.EOF) {
 				break
 			}
-			return 0, err
+
+			return 0, errors.WithStack(err)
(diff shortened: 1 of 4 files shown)


jhillyerd:
Excellent, thank you

dcormier:
Any interest in tagging a release for that? I think it'd be nice to get that out there.

jhillyerd:
Going to hold off for a bit because of the recent post to #140. I suspect we are going to need to roll the continuation parsing code back to a state where it just follows the RFCs.

PR #176, opened 2020-11-21
Correct shadowed errors

jhillyerd:
This is in preparation for introducing more thorough static checks on PRs.


dcormier:
It seems like the shadowing on those was inconsequential, since it looks like the `error` values was used immediately in those cases.

The static checks might be a little aggressive if it complained about those.

jhillyerd:
Yeah, it's definitely more aggressive than I'd like.  Haven't looked into tuning it, but given enmime is dealing with potentially hostile external email for many folks, probably want static checks.

PR #178, opened 2020-11-22
ParseMediaType: Don't split quoted strings containing semicolons

### LINKED ISSUE #174: Attachment having semicolon (;) in filenames are not downloading propery.
What I did: attach a file having name like abc;xyz.doc, test;file.txt

What I expected: File should be properly downloaded with complete name.

What I got: File names are coming partially, it is getting broken  at semicolon (;). So I am only getting file-names as abc, test etc. Also extension is not coming. So files are not opening.

Release or branch I am using: version->v0.8.2, commit id->2a0b399

(Please attach a sample message if you feel it will help reproduce the issue)

jhillyerd:
- [x] Check RFCs for interaction w/ quoting rules

Fixes #174


jhillyerd:
It looks like others are also unclear about contents and escaping of quoted strings: https://stackoverflow.com/questions/59816619/do-quoted-mime-media-type-parameters-allow-escaped-quotes

Given that Go let's us escape things with its mime library, I will try to support it as well.

https://play.golang.org/p/r_8sUV16lax

jhillyerd:
@requaos and @dcormier -- if either of you have time to review this PR, feedback appreciated.  Thanks!

header_test.go
```diff
@@ -728,3 +734,86 @@ func TestCommaDelimitedAddressLists(t *testing.T) {
...
+			mtype:  "text/html",
+			params: map[string]string{"name": "index;a.html", "hash": "8675309"},
+		},
+		{
+			label:  "quoted separator with escaped quote mid-string",
+			input:  "text/html; name=\"index\\\";a.html\"; hash=8675309",
+			mtype:  "text/html",
+			params: map[string]string{"name": "index\";a.html", "hash": "8675309"},
+		},
+	}
```
dcormier, line 793: The test cases, above, seem like what I would expect. I'd like @requaos to take a look with what he's seen in the wild, too.

internal/stringutil/split_test.go
```diff
@@ -0,0 +1,72 @@
...
+			want:  []string{"a", "b", "c"},
+		},
+		{
+			input: "a;'b;c';d",
+			want:  []string{"a", "'b;c'", "d"},
+		},
+		{
+			input: "a;'b;c;d",
+			want:  []string{"a", "'b;c;d"},
+		},
```
dcormier, line 44: I see you're thinking here (treat unterminated quote as if terminated at EOF), but I'm not sure if it's the right approach. I don't necessarily think it's the wrong approach, either. I'm just not sure.

requaos:
👀 today

internal/stringutil/split_test.go
jhillyerd: I can picture two possible alternatives; if we hit the end of the string and find it has an unterminated quote:

1. Start over and split the entire string without respecting separator quoting at all -- aka how enmime worked previously, since it didn't have a notion of quoting.
2. Reset to the last opening quote and pretend it didn't exist; so earlier quoted runs would be respected.

I'd lean towards the first option, given we don't know which quote(s) may be missing, the second isn't going to be any safer than what the code does now.

dcormier: I think I'd agree with you about going with the first option.

requaos:
So here is a nice quote from that SO thread:

> Figuring out what exactly the RFC stipulates is not going to be actually useful. The de facto standard you need to adhere to for interoperability is set by poorly implemented mail clients and browsers. If that's your goal, I would not expect anything like escaping to work reliably, regardless of whether the RFC is on your side.

So I looked up mozilla's libraries and found some stuff of interest: https://github.com/mozilla-comm/jsmime/blob/master/headerparser.js#L29
`Quoted-strings (if opt.qstring is true): A string which is surrounded by double quotes. Escapes in the string are omitted when returning.` 

Another note about this behavior: `[Whitespace] need not be between a qstring and an atom (a"b" produces two tokens, a and b). This is an error case, though.`

However, in the implementation I see a tell: This is at the head of the character iteration loop which is designed to detect quoted-strings and domain-literals.

```
    // If we see a \, no matter what context we are in, ignore the next
    // character.
    if (ch == '\\') {
      i++;
      continue;
    }
```

So it looks like the accepted best practice here is simply to omit any escaped characters, whether or not it's within a quoted-string.

I'm still looking for how they handle single versus double quotes inside each other, like the proposed test-case: `"a;'b;c;d"`

Ah, okay, so only double-quotes are respected as quoted-string delimiters, and single-quotes are treated as just another character.

So the `want` here should be:
```
		{
			input: "a;'b;c;d",
			want:  []string{"a", "'b", "c", "d"},
		},
```

requaos:
@jhillyerd I think that if a user wants to include a delimiter with a quoted-string, it must be done using RFC2047 encoding.
Such that a filename like:
```
name="Index\";a.doc"
```
is rendered as:
```
name="=?UTF-8?B?SW5kZXhcIjthLmRvYw==?="
```

jhillyerd:
For context, I only used the single quotes in my test cases to make them easier to read, they won't be respected as quoting characters in actual mail.  I'll switch to backtick string literals to help make that more clear.

jhillyerd:
https://tools.ietf.org/html/rfc5322#section-3.2.1 doesn't mention being able to quote a quoting character, Neil is probably right that this isn't something that should be done in MIME, so likely Go's mime headers just encode incorrectly.  I'll remove that from the splitter.

jhillyerd:
I think I've addressed everything, we probably won't behave identical to Mozilla's library, but close enough.  I'll merge this tomorrow if nobody has any add'l comments.

PR #179, opened 2020-11-22
add SendWithEnvelopFrom

cgroschupp:


jhillyerd:
Thanks for sending this PR, can you elaborate it's purpose?  Do you need the from value send to the SMTP server to be different from that of the email headers?

If you are just looking to have multiple from addresses:

> Each manipulation method **returns a copy** of the receiver struct.  It can be considered immutable if the caller does not
> modify the string and byte slices passed in.  Immutability allows the headers or entire message to be reused across multiple threads.

This means you can setup an email and send it with different from already:

```
msg := Builder()
// configure message

msg.From("Bob", "bob@example.com").Send(...)
msg.From("Fred", "fred@example.com").Send(...)
```

cgroschupp:
@jhillyerd Yes, I need a different address in the envelope. I don't know if the function name 'SendWithEnvelopFrom' is the best, maybe you have a better idea.

jhillyerd:
I feel using the term envelope here could be confusing since the parsing portion of the library has an Envelope struct.  It appears the RFCs use the term "return path" -- but that also seems a bit dated.  How about ReturnAddress?

jhillyerd:
Friendly ping on function rename

cgroschupp:
@jhillyerd sorry for the delay.

jhillyerd:
Thanks for this!

jhillyerd:
Heads up that this got renamed to `SendWithReversePath` in #182

Issue #158, opened 2020-07-01
Use envelope to create a pre filled message builder


dolanor:
What I did: use the lib and parsed an envelope, changed some `from:` and `to:` fields

What I expected: use the envelope as is to send it through smtp with the updated `from:` and `to:`

What I got: no API exists to get that modified envelope without building it from scratch with the Builder

Release or branch I am using: v0.8.1

jhillyerd:
The builder is a simplified subset of what's possible in MIME, so Builder can convert to Envelope, but not the other way around.  However, we could add a Send function to Envelope as well.

PR #138, opened 2019-11-23
Increase test coverage

requaos:
Just trying to get that test coverage number up

PR #182, opened 2021-02-14
Add an abstraction layer for sending messages

jhillyerd:
This PR abstracts the sending of messages via SMTP, allowing users to substitute their own sending routines.  It also lets enmime and its users to test the sending of mail via mocks or fakes.

First step for #158 

*Warning:* This does change the builder API, will require new semantic release.

### DIFF
--- a/builder.go
+++ b/builder.go
@@ -330,7 +330,8 @@ func (p MailBuilder) SendWithReturnAddress(sender Sender, from string) error {
 	return sender.Send(from, recips, buf.Bytes())
 }
 
-// Send encodes the message and sends it via the specified Sender.
+// Send encodes the message and sends it via the specified Sender, using the address provided to
+// `From()` as the return address.
 func (p MailBuilder) Send(sender Sender) error {
 	return p.SendWithReturnAddress(sender, p.from.Address)
 }
--- a/builder.go
+++ b/builder.go
@@ -306,8 +306,8 @@ func (p MailBuilder) Build() (*Part, error) {
 	return root, nil
 }
 
-// SendWithReturnAddress encodes the message and sends it via the specified Sender.
-func (p MailBuilder) SendWithReturnAddress(sender Sender, from string) error {
+// SendWithReversePath encodes the message and sends it via the specified Sender.
+func (p MailBuilder) SendWithReversePath(sender Sender, from string) error {
 	buf := &bytes.Buffer{}
 	root, err := p.Build()
 	if err != nil {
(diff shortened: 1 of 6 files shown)


requaos (approved):
We may want to bubble this excerpt up from the smtp package documentation about sending to BCC recipients:
```
// The msg headers should usually include
// fields such as "From", "To", "Subject", and "Cc".  Sending "Bcc"
// messages is accomplished by including an email address in the to
// parameter but not including it in the msg headers.
```
Otherwise LGTM

builder.go
```diff
@@ -329,13 +327,13 @@ func (p MailBuilder) SendWithReturnAddress(addr, from string, a smtp.Auth) error
...
-	return smtp.SendMail(addr, a, from, recips, buf.Bytes())
+	return sender.Send(from, recips, buf.Bytes())
 }
 
-// Send encodes the message and sends it via the SMTP server specified by addr. Send uses
-// net/smtp.SendMail, and accepts the same authentication parameters.
-func (p MailBuilder) Send(addr string, a smtp.Auth) error {
-	return p.SendWithReturnAddress(addr, p.from.Address, a)
+// Send encodes the message and sends it via the specified Sender, using the address provided to
+// `From()` as the return address.
```
dcormier, line 334:
```suggestion
// `From()` as the reverse-path.
```

"Return address" is a bit ambiguous. Is that the `Reply-To` header? Is that the [reverse-path](https://tools.ietf.org/html/rfc5321#section-3.3) (from the `MAIL FROM` command)?

sender_test.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+func TestSMTPSend(t *testing.T) {
+	from := "user@example.com"
+	auth := smtp.PlainAuth("", from, "password", "mail.example.com")
+	s := enmime.NewSMTP("0.0.0.0", auth)
+
+	// Satisfy requirements of the Send method and use an intentionally malformed From Address to
+	// elicit an expected error from smtp.SendMail, which can be type-checked and verified.
+	err := s.Send(from+"\rinvalid", []string{"to@example.com"}, []byte("message content"))
```
dcormier, line 18: Should also validate that `\n` is treated as an invalid character.

jhillyerd: This code isn't really interested in what is invalid, it's just testing that Go's built-in Send is actually called.  It's just a refactored version of what @requaos did in #138 -- bit of a hack to maximize test coverage without being able to mock.

builder.go
jhillyerd: Done, thanks

jhillyerd:
Added a reworked version of that Go Bcc comment, and renamed to/from to recipients/reversePath.

PR #183, opened 2021-03-03
Add failing test with an unquoted file name

Pritoj:
Currently the ParseMimeType function is unable to handle the case where
a param has spaces and is unquoted


Pritoj:
@jhillyerd I have an ugly fix that works for now. I will spend some time to clean up etc. Could you please help verify that this is acceptable?

kode4food:
@Pritoj, @jhillyerd,

I'm wondering if it makes more sense for this type of behavior to be exposed as a general-purpose header field filter? Something like:

```go
type HeaderFieldFilter = func(name string, body string) (name string, body string)
```

And then configured as optional encode and decode filters?

jhillyerd:
Overall this looks reasonable to me.  I worry a bit about performance and allocations, but this applies to a lot of the code in header.go.  Go ahead and cleanup, and then I will review.  For your two anonymous functions, can you check that [Go is inlining](https://gist.github.com/awalterschulze/1b3f421dd3c71bad2e9a135d8216b2a4) them?  If they are not inlined, I don't think the readability improvement is worth the cost.

Likely needs more tests for edge cases.

@kode4food I don't think many users are going to understand what all these header fixes do, or when they'd need them, which is part of the reason I haven't added options to enmime yet.  One advantage of moving it an interface similar to yours would be not having to allocate string builders & buffers for each header line though.  I don't want to tackle any of this before 1.0 release, as I feel enmime is very close.

jhillyerd:
Obsoleted by #209

Issue #186, opened 2021-03-30
MailBuilder Validation for To/CC/BCC()


jawr:
MailBuilder.Build() does some validation around To/CC/BCC here:

https://github.com/jhillyerd/enmime/blob/master/builder.go#L221

However, calling MailBuilder.To/CC/BCC() with a blank address appends an empty struct which then circumvents the validation. Should there be some sort validation on To/CC/BCC() to prevent this, or is it the callers responsibility?

jhillyerd:
If we can do that validation inexpensively, then I'm in favor of adding it.

jawr:
My immediate thought is to just check for length and maybe '@' on the addr string. I don't think it's enmimes job to do much more than prevent emails with a blank To going out. If that's agreeable I'll make a PR.

jhillyerd:
Addresses don't always need an @ sign, so we shouldn't enforce that.  Length sounds good to me, thanks.

PR #187, opened 2021-04-01
Validate To/BCC/CC Ref #186

jawr:
References  #186

Only append To/BCC/CC if they have an address with a length. Added some tests.

The tests are slightly verbose as I couldn't think of a nice way of tabulating them, plus having the error message hard coded instead of as a referenceable error value feels strange.

### DIFF
--- a/builder_test.go
+++ b/builder_test.go
@@ -1005,64 +1005,61 @@ func TestSendWithReversePath(t *testing.T) {
 }
 
 func TestEmptyTo(t *testing.T) {
-	sender := &mockSender{}
 	from := "from@example.com"
 	text := []byte("test text body")
-	ret := "return@example.com"
+	rcpt := "rcpt@example.com"
 	a := enmime.Builder().
 		Text(text).
 		From("name", from).
 		Subject("foo").
-		To("", "")
+		To(rcpt, "")
 
-	err := a.SendWithReversePath(sender, ret)
+	_, err := a.Build()
 	if err == nil {
 		t.Fatal("Expected error, got nil")
 	}
 
-	if err.Error() != "no recipients (to, cc, bcc) set" {
-		t.Fatalf("Unexpected error, wanted 'no recipients (to, cc, bcc) set' got %s", err)
+	if err.Error() != enmime.ErrorMissingRecipient {
+		t.Fatalf("Unexpected error, wanted %q got %s", enmime.ErrorMissingRecipient, err)
 	}
 }
 
 func TestEmptyBCC(t *testing.T) {
-	sender := &mockSender{}
 	from := "from@example.com"
 	text := []byte("test text body")
-	ret := "return@example.com"
+	rcpt := "rcpt@example.com"
 	a := enmime.Builder().
 		Text(text).
 		From("name", from).
 		Subject("foo").
-		BCC("", "")
+		BCC(rcpt, "")
 
-	err := a.SendWithReversePath(sender, ret)
+	_, err := a.Build()
 	if err == nil {
 		t.Fatal("Expected error, got nil")
 	}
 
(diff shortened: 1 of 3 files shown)


builder_test.go
```diff
@@ -1003,3 +1003,66 @@ func TestSendWithReversePath(t *testing.T) {
...
+	from := "from@example.com"
+	text := []byte("test text body")
+	ret := "return@example.com"
+	a := enmime.Builder().
+		Text(text).
+		From("name", from).
+		Subject("foo").
+		To("", "")
+
+	err := a.SendWithReversePath(sender, ret)
```
jhillyerd, line 1018: Let's test with Build, instead of Send

builder_test.go
```diff
@@ -1003,3 +1003,66 @@ func TestSendWithReversePath(t *testing.T) {
...
+	if err == nil {
+		t.Fatal("Expected error, got nil")
+	}
+
+	if err.Error() != "no recipients (to, cc, bcc) set" {
+		t.Fatalf("Unexpected error, wanted 'no recipients (to, cc, bcc) set' got %s", err)
```
jhillyerd, line 1024: We could add the text of the error to https://github.com/jhillyerd/enmime/blob/master/error.go and use it in both places.

builder_test.go
```diff
@@ -1003,3 +1003,66 @@ func TestSendWithReversePath(t *testing.T) {
...
+func TestEmptyTo(t *testing.T) {
+	sender := &mockSender{}
+	from := "from@example.com"
+	text := []byte("test text body")
+	ret := "return@example.com"
+	a := enmime.Builder().
+		Text(text).
+		From("name", from).
+		Subject("foo").
+		To("", "")
```
jhillyerd, line 1016: Let's put a value in the name section to prove we are testing for emtpy addr only.

jhillyerd (changes requested):

builder_test.go
```diff
@@ -1005,64 +1005,61 @@ func TestSendWithReversePath(t *testing.T) {
 }
 
 func TestEmptyTo(t *testing.T) {
-	sender := &mockSender{}
 	from := "from@example.com"
 	text := []byte("test text body")
-	ret := "return@example.com"
+	rcpt := "rcpt@example.com"
```
jhillyerd, line 1010:
```suggestion
	rcpt := "rcpt name"
```

builder_test.go
```diff
@@ -1005,64 +1005,61 @@ func TestSendWithReversePath(t *testing.T) {
...
+		t.Fatalf("Unexpected error, wanted %q got %s", enmime.ErrorMissingRecipient, err)
 	}
 }
 
 func TestEmptyBCC(t *testing.T) {
-	sender := &mockSender{}
 	from := "from@example.com"
 	text := []byte("test text body")
-	ret := "return@example.com"
+	rcpt := "rcpt@example.com"
```
jhillyerd, line 1030:
```suggestion
	rcpt := "rcpt name"
```

builder_test.go
```diff
@@ -1005,64 +1005,61 @@ func TestSendWithReversePath(t *testing.T) {
...
+		t.Fatalf("Unexpected error, wanted %q got %s", enmime.ErrorMissingRecipient, err)
 	}
 }
 
 func TestEmptyCC(t *testing.T) {
-	sender := &mockSender{}
 	from := "from@example.com"
 	text := []byte("test text body")
-	ret := "return@example.com"
+	rcpt := "rcpt@example.com"
```
jhillyerd, line 1050:
```suggestion
	rcpt := "rcpt name"
```

jhillyerd (changes requested):
one small tweak, otherwise looks great, thanks

jhillyerd:
Thank you!

Issue #188, opened 2021-04-02
Test coverage action failing


jhillyerd:
Our Test Coverage Github action errors

```
/home/runner/work/_actions/shogo82148/actions-goveralls/v1/bin/goveralls_linux_amd64 -service=github
cannot find file "github.com/jhillyerd/enmime/boundary.go": can't find "boundary.go": go/build: go list github.com/jhillyerd/enmime/: exit status 2
go: cannot find GOROOT directory: /usr/local/go


Error: The process '/home/runner/work/_actions/shogo82148/actions-goveralls/v1/bin/goveralls_linux_amd64' failed with exit code 1
```

jhillyerd:
May have been an external issue, since the old run passes here: https://github.com/jhillyerd/enmime/runs/2263116228?check_suite_focus=true

But the parallel approach is likely better.

PR #189, opened 2021-04-04
Issue 188 Failing Github Actions

jawr:


jawr:
Ref #188 

Not sure why it didn't run the build-and-test action here. I checked and saw that your old method used to work, but looking at the documentation and having played with this recently on another repo I couldn't work out where the profile was being made.

If this is not how you wanted to do it feel free to close.

jhillyerd:
I think the way I had it configured previously, it ran the unit tests again and captured the coverage then, instead of collecting coverage from already run tests.  I'll try merging this into a branch and see what happens.

jhillyerd:
`You have an error in your yaml syntax on line 22` yaml sucks

PR #193, opened 2021-05-28
fix split when spaces are around ';'

gopher-johns:
I had some issues with e-mail client adding some spaces in content-type.
here is an example of faulty content-type I found
```Content-Type: multipart/mixed; boundary="DELIMITERMAIL" ; charset="utf-8"```

It cause a bad mime type parsing error.

I not sure that I choose the best spot to fix that.

### CODE REVIEW

**Comment by jhillyerd** [comment]
Thanks for the PR.  As you suspected, this is probably not the right place for it.  It's also meant to be a tight character based loop, so bringing in the strings package isn't really appropriate.

I'd start by adding a failing test case to https://github.com/jhillyerd/enmime/blob/master/header_test.go#L753 and then see if you can fix it in header.go

**Comment by demofrager** [comment]
Hi, do you think this could be the same issue? 
```
...
--C90E372A1E49414A9DA1EC64032CE50C
Content-Type: text/plain; charset="UTF-8"
	; format=flowed
```

This would be the header test that breaks the ParseMediaType (I also added yours): 
```golang 
		{       //gopher-johns
			label:  "with space",
			input:  `text/plain; charset="UTF-8" ; format=flowed`,
			mtype:  "text/plain",
			params: map[string]string{"charset": "UTF-8", "format": "flowed"},
		},
		{
			label: "with tabs and new lines",
			input: `text/plain; charset="UTF-8"
			; format=flowed`,
			mtype:  "text/plain",
			params: map[string]string{"charset": "UTF-8", "format": "flowed"},
		},
```
Sorry if this doesn't belong to the same issue. I'll open a new one and try to provide a fix.

Thanks for the package :)

jhillyerd:
They certainly look related to me.

jhillyerd:
This was fixed in #195

PR #195, opened 2021-06-08
Remove whitespace between parameter=value and ; in headers

demofrager:
This is a fix for the problem of PR: [193](https://github.com/jhillyerd/enmime/pull/193)

Sorry for taking over, but I really need the fix :(

Let me know if there is anything you would like me to change.

Thanks for the package.

### DIFF
--- a/header.go
+++ b/header.go
@@ -309,9 +309,7 @@ func fixMangledMediaType(mtype string, sep rune) string {
 		// semicolon.
 		if i != len(parts)-1 && !strings.HasSuffix(mtype, ";") {
 			// Remove whitespace between parameter=value and ;
-			for strings.HasSuffix(mtype, " ") || strings.HasSuffix(mtype, "\t") {
-				mtype = mtype[:len(mtype)-1]
-			}
+			mtype = strings.TrimRight(mtype, " \t")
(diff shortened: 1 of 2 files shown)


header.go
```diff
@@ -308,6 +308,10 @@ func fixMangledMediaType(mtype string, sep rune) string {
 		// Only terminate with semicolon if not the last parameter and if it doesn't already have a
 		// semicolon.
 		if i != len(parts)-1 && !strings.HasSuffix(mtype, ";") {
+			// Remove whitespace between parameter=value and ;
+			for strings.HasSuffix(mtype, " ") || strings.HasSuffix(mtype, "\t") {
```
jhillyerd, line 312: https://staticcheck.io/docs/checks#S1017 suggests we replace this check and slice with two calls to strings.TrimSuffix

I also wonder if it would be better to use TrimRight, as your implementation will only remove a single trailing whitespace character.

demofrager: I think TrimRight is a better approach. It removes everything in the cutset. No need for 2 TrimSuffix calls.

jhillyerd:
Looks good, thanks!

PR #199, opened 2021-06-28
Empty content type parameter ignored

pavelbazika:
There are some emails which contain slighlty malformed Content-Type headers in form
```
Content-Type: application/x-pdf; ; name="=?utf-8?Q?PR-8469=5F2018.pdf?="
```
Function `fixMangledMediaType` turns such header to
```
Content-Type: application/x-pdf; =not-a-param-value; name="=?utf-8?Q?PR-8469=5F2018.pdf?="
```
This patch ignores empty parameter instead. According to RFC 1521 "the ordering of parameters is not significant" thus it should not harm anything.

### DIFF
--- a/header.go
+++ b/header.go
@@ -284,6 +284,11 @@ func fixMangledMediaType(mtype string, sep rune) string {
 				continue
 			}
 
+			if len(strings.TrimSpace(p)) == 0 {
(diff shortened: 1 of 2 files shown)


jhillyerd:
Nice catch, thanks.

PR #200, opened 2021-06-30
Boundary reader - long lines processing fixed

pavelbazika:
Hi,

when there is a line longer than 4096 bytes, message parsing fails, because boundary reader returns bufio.ErrBufferFull.

This patch handles reading longer lines than 4096 bytes.

### DIFF
--- a/boundary.go
+++ b/boundary.go
@@ -181,10 +181,20 @@ func (b *boundaryReader) Next() (bool, error) {
 		_, _ = io.Copy(ioutil.Discard, b)
 	}
 	for {
-		line, err := b.r.ReadSlice('\n')
-		if err != nil && err != io.EOF {
-			return false, errors.WithStack(err)
+		var line []byte
+		var err error
+		for {
+			// Read whole line, handle extra long lines in cycle
+			var segment []byte
+			segment, err = b.r.ReadSlice('\n')
+			line = append(line, segment...)
+			if err == nil || err == io.EOF {
+				break
+			} else if err != bufio.ErrBufferFull || len(segment) == 0 {
+				return false, errors.WithStack(err)
+			}
 		}
+
 		if len(line) > 0 && (line[0] == '\r' || line[0] == '\n') {
 			// Blank line
 			continue
--- a/boundary.go
+++ b/boundary.go
@@ -181,13 +181,18 @@ func (b *boundaryReader) Next() (bool, error) {
 		_, _ = io.Copy(ioutil.Discard, b)
 	}
 	for {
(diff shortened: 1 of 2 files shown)


boundary.go
```diff
@@ -181,10 +181,20 @@ func (b *boundaryReader) Next() (bool, error) {
...
-		line, err := b.r.ReadSlice('\n')
-		if err != nil && err != io.EOF {
-			return false, errors.WithStack(err)
+		var line []byte
+		var err error
+		for {
+			// Read whole line, handle extra long lines in cycle
+			var segment []byte
+			segment, err = b.r.ReadSlice('\n')
+			line = append(line, segment...)
```
jhillyerd, line 190: A concern I have here is we are going to be doing a lot more allocations than we used to.  In the previous version of this loop, `line` was just a pointer into the bufio buffer, but now we are allocating and growing `line` multiple times per Next().

Is it feasible to move the line buffer into a `b` struct and re-use it?

boundary_test.go
```diff
@@ -477,6 +477,24 @@ func TestBoundaryReaderReadErrors(t *testing.T) {
 	}
 }
 
+func TestBoundaryReaderLongLine(t *testing.T) {
```
jhillyerd, line 480:
```suggestion
// TestBoundaryReaderLongLine checks that boundaryReader can read lines longer than the `peekBufferSize`.
func TestBoundaryReaderLongLine(t *testing.T) {
```

boundary_test.go
```diff
@@ -477,6 +477,24 @@ func TestBoundaryReaderReadErrors(t *testing.T) {
...
+	//dest := make([]byte, 8 *1024)
```
jhillyerd, line 481: remove

jhillyerd (changes requested):
Nice find.

boundary.go
pavelbazika: What about allocating only when the line is longer than buffer?

In fact, this is not a bug at all. MIME mail must have at most 998 octects on the line according to RFC, so it's about tolerance of some existing malformed emails.

boundary_test.go
pavelbazika: Comment added, thanks.

boundary_test.go
pavelbazika: Oops, removed. I can squash the fixes and force push as a single commit if you want. I'm not sure if github can do this automatically like gitlab.

jhillyerd:
I like the new approach, thanks.

PR #220, opened 2021-09-29
Fixed crash in QPCleaner, when line is too long and buffer is almost full

pavelbazika:
Hi,

there is a bug in new QPCleaner implementation, which causes panic. When the `qp.lineLen` exceeds `MaxQPLineLen`, newline is safely inserted to `dest` buffer. But it is not checked, if `n` is still `< destLen` after this insertion, causing write after the end of `dest` in the following `switch` statement.

Here is a fix and test, the test causes panic on current version, the checks are there just for completeness.

Best regards

Pavel


jhillyerd:
Thanks for catching this!

PR #224, opened 2022-01-03
Fix newlines in mediatype.Parse

ungerik:
Added a very simple `fixNewlines` function to handle the following outlook produced case: 

```
		{
			label:  "Outlook-Logo with newlines",
			input:  `application/octet-stream; name="=?utf-8?B?T3V0bG9vay1Mb2dvCgpEZXNj?="`,
			mtype:  "application/octet-stream",
			params: map[string]string{"name": "Outlook-Logo  Desc"},
		},
```

### DIFF
--- a/mediatype/mediatype.go
+++ b/mediatype/mediatype.go
@@ -39,9 +39,10 @@ const (
 //   * Missing ';' between content-type and media parameters
 //   * Repeating media parameters
 //   * Unquoted values in media parameters containing 'tspecials' characters
+//   * Newline characters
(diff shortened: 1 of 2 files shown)


jhillyerd:
While I agree that I wouldn't want a newline in a filename, I'm not sure this is something that should be fixed in enmime, since clients can choose how to handle newlines themselves?  I've tried to add well meaning things like this to enmime in the past and caused problems for others.

ungerik:
I wouldn't mind the newlines, but the go std package refuses to parse values with newlines.

jhillyerd:
That sounds like a good reason to me, thanks.

PR #225, opened 2022-01-17
fix(fixMangledMediaType) remove extra parts

gopher-johns:
Hello,

I encountered some unknown email client in production. It used an invalid content type with extra parts, like `application/pdf/.pdf`

I `fixed` parsing by removing extra parts, but I'm not 100% sure it could be split as 
* `application` `pdf`
* `application` `pdf/.pdf`
* `application` `pdf-.pdf`

Thank you again for your great work


jhillyerd:
I think this is a good start, and if folks run into issues we can try to tune it.

PR #229, opened 2022-01-27
Remove <!DOCTYPE tag in bad content types

csucu:
We came across a bad content type, which contained a doctype tag. This Change ignores it.


jhillyerd:
I feel this is fixup is overly specific; it may make more sense to strip out any HTML tag, not just doctype.

But more concerning to me is that the Index into the string does not respect quoted values, which could further mangle a properly quoted string.

jhillyerd:
A unit test to show what happens when the match is inside a string may be enough to prove we don't do too much harm, I'm not certain.

jhillyerd:
Thanks, much appreciated!

Issue #218, opened 2021-09-20
Failed to parse address with colon


lcd1232:
What I did: I have valid header `From: =?utf-8?Q?re=3AStore=20?=
 =?utf-8?Q?Apple=20?=
 =?utf-8?Q?Premium=20?=
 =?utf-8?Q?Reseller?= <noreply@mail.restore.ru>`

What I expected: Parsed successfully

What I got:  error = `mail: expected comma`

Release or branch I am using: master

Comment: looks like we need to add quotes as it [standard lib](https://play.golang.org/p/12Njee_-HSB) do

Example message:
```
Received: id 2D333CD836-1631660949
Message-Id: <20210915020909.0.20210915020909_pe_@CmKAw7AV7BGnRfumzamcPQ3185481631660948.escheclub.pegate>
Date: Wed, 15 Sep 2021 02:09:09 +0300
From: =?utf-8?Q?re=3AStore=20?=
 =?utf-8?Q?Apple=20?=
 =?utf-8?Q?Premium=20?=
 =?utf-8?Q?Reseller?= <noreply@mail.restore.ru>
To:  <user@example.com>
Subject: =?utf-8?Q?iPhone=20?=
 =?utf-8?Q?13=20?=
 =?utf-8?Q?Pro=2E=20?=
 =?utf-8?Q?=D0=9D=D0=BE=D0=B2=D1=8B=D0=B5=20?=
 =?utf-8?Q?=D0=BA=D0=B0=D0=BC=D0=B5=D1=80=D1=8B=2E=20?=
 =?utf-8?Q?=D0=9D=D0=B5=D1=80=D0=B5=D0=B0=D0=BB=D1=8C=D0=BD=D1=8B=D0=B5=20?=
 =?utf-8?Q?=D0=B2=D0=BE=D0=B7=D0=BC=D0=BE=D0=B6=D0=BD=D0=BE=D1=81=D1=82=D0=B8=2E?=

```

jhillyerd:
Sorry for incredibly slow response on this, motivation has been scarce lately.  :/  I spent a bit of time playing around with this and your #219 today.

I'm wondering if we need to take a step back, and figure out how much of our AddressList function is actually needed anymore. I don't think Go (properly?) decoded rfc2047 encoded headers back when it was originally committed in https://github.com/jhillyerd/enmime/commit/a09ca5ec5882e3fe8faee0d7d40cb774f5272c8c

[...]

PR #237, opened 2022-03-03
fix: fix for quote-printed utf-8 header with quotes

dmytrokasianenko-outreach:
This PR fixes decoding addresses with quotes. 
For example, `"=?UTF-8?Q?Miros=C5=82aw_Marczak?="` should be decoded as `Mirosław Marczak`.

### DIFF
--- a/envelope_test.go
+++ b/envelope_test.go
@@ -2,6 +2,8 @@ package enmime_test
 
 import (
 	"bytes"
+	"fmt"
+	"net/mail"
 	"sort"
 	"strings"
 	"testing"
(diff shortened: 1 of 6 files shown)


jhillyerd:
I think this may revert the fix for #218 by decoding the 2047 words too early.

jhillyerd:
Thinking about how to get better test coverage for this, without introducing a bunch more file based envelope tests: 

We could copy https://github.com/jhillyerd/enmime/blob/master/header_test.go#L60 to envelope_test.go, then modify it to create a minimal message, putting the input values on the `To:` line

dmytrokasianenko-outreach:
You are right, it was breaking the the fix for https://github.com/jhillyerd/enmime/issues/218.

I've changed solution by trying to additionally decode `address.Name` with `mime.WordDecoder`. It works fine for the test cases, however I'm not sure it is a valid one. What do you think?

dmytrokasianenko-outreach:
I've added a fix for `ParseAddressList` to be able to decode different charsets

jhillyerd:
I'll try and take a deeper look at the address.Name decoding in the next few days, thanks

jhillyerd:
I think this is a good, thanks for your work on this!

dmytrokasianenko-outreach:
thank you for all your work!

PR #238, opened 2022-03-14
fix: parse address joined with semicolons

dmytrokasianenko-outreach:
Fix for semicolons in address. Should handle addresses joined by semicolon and remove semicolons where necessary


internal/stringutil/addr.go
```diff
@@ -60,8 +60,13 @@ func EnsureCommaDelimitedAddresses(s string) string {
 			}
 			if inDomain {
 				if r == ';' {
-					sb.WriteRune(r)
-					break
+					inDomain = false
+					if i == len(s)-1 { // if last one
```
jhillyerd, line 64:
```suggestion
					if i == len(s)-1 {
						// Preserve trailing semicolon.
```

jhillyerd (changes requested):
One minor comment nitpick.

Can we also add a semicolon separated list or two to https://github.com/jhillyerd/enmime/blob/master/envelope_test.go#L761 ?

internal/stringutil/addr.go
dmytrokasianenko-outreach: It won't preserve trailing semicolon, it will omit it.
I've updated comment

dmytrokasianenko-outreach:
> One minor comment nitpick.
> 
> Can we also add a semicolon separated list or two to https://github.com/jhillyerd/enmime/blob/master/envelope_test.go#L761 ?

Added 🙂

PR #242, opened 2022-04-13
go 1.18 + fuzzing

lcd1232:


lcd1232:
@jhillyerd looks like the problem with grandcolline/golang-github-actions https://github.com/grandcolline/golang-github-actions/issues/20
We can switch to golangci-lint, see #243

PR #243, opened 2022-04-13
golangci-lint

lcd1232:
https://golangci-lint.run/


jhillyerd:
Thanks, I've heard good things about this one.  Hopefully it doesn't find too many problems with the existing codebase. :)

PR #244, opened 2022-04-14
Builder: Support other parts

### LINKED ISSUE #223: MailBuilder: support OtherParts
What I did: Tried to add OtherParts to Builder

What I expected: it's possible

What I got: there is no way to do it

Release or branch I am using: `github.com/jhillyerd/enmime v0.8.4`

lcd1232:
Closes #223
I moved other parts in multipart/related as inlines because I saw [this image](https://stackoverflow.com/a/66551704/5996922)
Tested on eml message, looks fine


jhillyerd:
I actually removed testify from the codebase back in 2016, would rather not re-introduce it, not a fan.  https://github.com/jhillyerd/enmime/commit/8f740ea94617eb982c13213ab4ca41ff05186a84

iredmail:
I like testify since it saves a lot typing. :)

jhillyerd:
It was a while ago, so I forget the exact reasons, but my primary problem was that it's test output was not parseable when running the tests from vim, the colors looked like crap and jumping to the point of failure didn't work.   I also didn't like importing unnecessary things that could cause conflicts as a library to be used by others, I think go mod is smart enough to not force test dependencies on our users so probably less of a problem nowdays.

lcd1232:
@jhillyerd do you think I should remove testify?

jhillyerd:
I tried this morning, testify no longer breaks go-vim, and doesn't seem any worse than any other failed tests with multi-line output.

If I read https://go.dev/ref/mod#graph-pruning correctly, our choices will still impact the modules that import us until everyone is on Go 1.18.

With that in mind, let's keep testify, but switch from 1.7.1 (very new, and not properly released on github), to 1.7.0 which has been around for more than a year.  That way folks that import us have more control over when they move to 1.7.1.

go.mod
```diff
@@ -10,6 +10,7 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/rivo/uniseg v0.2.0 // indirect
 	github.com/ssor/bom v0.0.0-20170718123548-6386211fdfcf // indirect
+	github.com/stretchr/testify v1.7.1 // indirect
```
jhillyerd, line 13: Switch to 1.7.0

jhillyerd (changes requested):

go.mod
lcd1232: Done

jhillyerd: Thanks!

PR #245, opened 2022-04-16
Fix linter issues

### LINKED ISSUE #241: PR checks fail
https://github.com/jhillyerd/enmime/blob/master/.github/workflows/pull-checks.yml is trying to use go 1.17 packages on 1.13 and failing

lcd1232:
Fixes #241

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -36,9 +36,13 @@ func (p *Part) Encode(writer io.Writer) error {
 	cte := p.setupMIMEHeaders()
 	// Encode this part.
 	b := bufio.NewWriter(writer)
-	p.encodeHeader(b)
+	if err := p.encodeHeader(b); err != nil {
+		return err
+	}
 	if len(p.Content) > 0 {
-		b.Write(crnl)
+		if _, err := b.Write(crnl); err != nil {
+			return err
+		}
 		if err := p.encodeContent(b, cte); err != nil {
 			return err
 		}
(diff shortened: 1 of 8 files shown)


header.go
```diff
@@ -21,13 +21,13 @@ const (
 	cdInline     = "inline"
 
 	// Standard MIME content types
-	ctAppPrefix        = "application/"
+	ctAppPrefix        = "application/" //nolint
```
lcd1232, line 24: @jhillyerd I need your decision on what to do with the unused code.

encode.go
```diff
@@ -36,9 +36,13 @@ func (p *Part) Encode(writer io.Writer) error {
 	cte := p.setupMIMEHeaders()
 	// Encode this part.
 	b := bufio.NewWriter(writer)
-	p.encodeHeader(b)
+	if err := p.encodeHeader(b); err != nil {
+		return err
+	}
 	if len(p.Content) > 0 {
-		b.Write(crnl)
+		if _, err := b.Write(crnl); err != nil {
```
jhillyerd, line 43: for all of the b.Writes in this func, it doesn't matter if they fail, as it will cause the b.Flush at the end to fail.  can we do a `_, _ :=` instead to remove the iferr noise?  Or disable linting in this part?  It makes it really hard to read.

header.go
jhillyerd: I think we can just delete these, seems unlikely we will use them.

jhillyerd (changes requested):

header.go
lcd1232: Done

encode.go
lcd1232: I'm not sure that we can ignore these errors because `bufio.NewWriter` can be changed in the future and `b.Flush` will not check previous errors. What do you think?

jhillyerd: That's not correct, because Flush() does not exist on other writers, so it will be obvious if the writer type were to be changed.

jhillyerd: Will resolve this for now, I'd like to submit this PR so that other PRs don't have to deal with the linter errors.

jhillyerd (approved):

jhillyerd:
Thanks!

PR #246, opened 2022-04-21
fix: parse media types which are escaped at first rune

dmytrokasianenko-outreach:
This commit fixing parsing media type params which stars with escaping.
Fox example `\"pdf\"` should be fixed as `"\"pdf\""`.


jhillyerd:
Thanks!

PR #247, opened 2022-04-22
fix: add decoding mail address

dmytrokasianenko-outreach:
This commit adds additional decoding for encoded email address.
For example: `=?UTF-8?B?dGVzdG5hbWU=?=@example.com` -> `testname@example.com`


jhillyerd:
Looks good, thanks!

PR #254, opened 2022-06-02
fix: buffer overrun when quoted printable is at the end of line and output buffer

pavelbazika:
In QPCleaner, write behind the buffer end could occur, when there was a `=XX` quoted printable token, line length was 1023 and remaining dest buffer space was only 1 byte. Because of line length, QPCleaner inserted `=\r\n`, but only the equal sign fit into the output buffer. The rest went to qp.overflow. 

Then the algorithm wrote to `dest[n]` without any additional checks, but `n` already pointed behind the buffer end.

Now the equal sign is also written to overflow in such situation.

I've also noticed, that qp.lineLength is not incremented when qp.overflow is put at the beginning of new buffer and also when `=` from quoted printable token is inserted. Seems as a bug to me, but I don't fix it in this PR.

### DIFF
--- a/internal/coding/quotedprint_test.go
+++ b/internal/coding/quotedprint_test.go
@@ -152,6 +152,40 @@ func TestQPCleanerLineBreakBufferFull(t *testing.T) {
 	}
 }
 
+func TestQPCleanerEqualSignOverflow(t *testing.T) {
+	// Test processing of =3D quoted printable token being at the end of line
+	// in combination with almost full buffer.
+	// qp cleaner should add =\r\n and then output =3D without buffer overflow
+
+	input := append(bytes.Repeat([]byte("abc"), 341), []byte("=3D")...)
+	inbuf := bytes.NewBuffer(input)
+	qp := coding.NewQPCleaner(inbuf)
+
+	output := make([]byte, 1024)
+	n, err := qp.Read(output)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if n != 1024 {
+		t.Errorf("Unexpected result length: %d", n)
+	}
+	if string(output[1020:]) != "abc=" {
+		t.Errorf("Unexpected result content: %s", output[1020:])
+	}
+
+	n, err = qp.Read(output)
+	if err != io.EOF {
+		t.Fatal(err)
+	}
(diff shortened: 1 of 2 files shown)


internal/coding/quotedprint_test.go
```diff
@@ -152,6 +152,40 @@ func TestQPCleanerLineBreakBufferFull(t *testing.T) {
...
+	inbuf := bytes.NewBuffer(input)
+	qp := coding.NewQPCleaner(inbuf)
+
+	output := make([]byte, 1024)
+	n, err := qp.Read(output)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if n != 1024 {
+		t.Errorf("Unexpected result length: %d", n)
```
jhillyerd, line 170: Can we add std Go "got: %v, wanted: %v" syntax to all four of these failures, they will be easily confused.

jhillyerd (changes requested):
Thanks, one small change requested.  Should I file a bug for the other issue you mentioned?

internal/coding/quotedprint_test.go
pavelbazika: Ok, done. I did not know about such std, personally I use `require` package.

pavelbazika:
> Thanks, one small change requested. Should I file a bug for the other issue you mentioned?

No need to file a bug, I'll fix it and come with a PR. I just wanted to ensure, that `lineLength` should be incremented in all circumstances.

internal/coding/quotedprint_test.go
jhillyerd: Thanks!

Not super important, but if you look through the go std lib, you will see this pattern over and over:

	if got, want := a.Get(&v1), 1; got != want {
		panic(fmt.Sprintf("Get() == %d, want %d", got, want))
	}

Basically they almost always use `got` and `want`, and got is placed before want in both the equality test and the failure output text.

We did introduce stretchr/testify back into enmime recently.  I wouldn't want to add `require` as well, but feel free to use testify assertions if you prefer those to vanilla go.

jhillyerd (approved):

jhillyerd:
Just tagged 0.10.0

PR #255, opened 2022-06-07
fix: QPCleaner line length equal sign counting

pavelbazika:
Hi,

as mentioned in https://github.com/jhillyerd/enmime/pull/254, this fixes missing counting of `=` when quoted printable encoded char is processed.

I've also mentioned, that line length is not incremented, when `qp.overflow` is put into the buffer at the beginning of the next `Read`. Tests show it is not needed, because `qp.lineLength` is set, when `qp.overflow` is filled at the end of previous `Read`.

Regards


jhillyerd:
Thank you!

PR #256, opened 2022-06-09
fix(mediatype): quoted semicolon

dmytrokasianenko-outreach:
This PR fixes splitting of Content-Type header, when parameter has semicolon inside quoted text`;`, e.g. `multipart/mixed; boundary="aaa;bbb;ccc"`. Previously, it was splitting wrongly, causing `fixUnescapedQuotes` function to break Content-Type header, making it `multipart/mixed; boundary=""`

Introduce new `FindQuoted` function. Reasoning: because for the splitting we still need to go through whole string, we can as well find all points where we need split. And then, use it for `IndexQuoted` or `CountQuoted` (analog of `strings.Index` and `strings.Count`), if we'll need it the future.

### DIFF
--- a/internal/stringutil/find.go
+++ b/internal/stringutil/find.go
@@ -1,8 +1,8 @@
 package stringutil
 
-// FindQuoted returns the indexes of the instance of v in s, or empty slice if v is not present in s.
+// FindUnquoted returns the indexes of the instance of v in s, or empty slice if v is not present in s.
 // It ignores v present inside quoted runs.
-func FindQuoted(s string, v rune, quote rune) []int {
+func FindUnquoted(s string, v rune, quote rune) []int {
 	escaped := false
 	quoted := false
 	indexes := make([]int, 0)
--- a/internal/stringutil/find.go
+++ b/internal/stringutil/find.go
@@ -1,0 +1,39 @@
+package stringutil
+
+// FindQuoted returns the indexes of the instance of v in s, or empty slice if v is not present in s.
+// It ignores v present inside quoted runs.
+func FindQuoted(s string, v rune, quote rune) []int {
+	escaped := false
+	quoted := false
+	indexes := make([]int, 0)
+	quotedIndexes := make([]int, 0)
+
+	for i := 0; i < len(s); i++ {
+		switch rune(s[i]) {
+		case escape:
+			escaped = !escaped // escape can escape itself.
+		case quote:
+			if escaped {
+				escaped = false
+				continue
+			}
+
+			quoted = !quoted
+			if !quoted {
+				quotedIndexes = quotedIndexes[:0] // drop possible indices inside quoted segment
+			}
+		case v:
+			escaped = false
+			if quoted {
+				quotedIndexes = append(quotedIndexes, i)
+			} else {
+				indexes = append(indexes, i)
+			}
+		default:
+			escaped = false
+		}
+
+	}
+
+	return append(indexes, quotedIndexes...)
+}
--- a/internal/stringutil/split_test.go
+++ b/internal/stringutil/split_test.go
@@ -4,7 +4,7 @@ import (
 	"testing"
 )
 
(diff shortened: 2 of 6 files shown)


internal/stringutil/find.go
```diff
@@ -0,0 +1,39 @@
+package stringutil
+
+// FindQuoted returns the indexes of the instance of v in s, or empty slice if v is not present in s.
+// It ignores v present inside quoted runs.
+func FindQuoted(s string, v rune, quote rune) []int {
```
jhillyerd, line 5: If I am understanding it's operation correctly, should this be called FindUnquoted?   Similar comment on the `quotedIndices` var name.

internal/stringutil/split_test.go
```diff
@@ -45,27 +45,102 @@ func TestSplitQuoted(t *testing.T) {
...
 		},
 		{
-			// Unterminated quoted-run will cause quotes to be ignored from the start of the string.
 			input: `"a;b";"c;d`,
-			want:  []string{`"a`, `b"`, `"c`, `d`},
+			want:  []string{`"a;b"`, `"c`, `d`},
 		},
 		{
 			// Quotes must be escaped via RFC2047 encoding, not just a backslash.
 			// b through c below must not be treated as a single quoted-run.
```
jhillyerd, line 53: I think this comment is now out of date, as we are treating b-c as single quoted run.  I am a little concerned that we are changing something that is noted as "must" in the comment, but given mediatype tests are still passing, I guess we will roll with it.

internal/stringutil/split_test.go
```diff
@@ -45,27 +45,102 @@ func TestSplitQuoted(t *testing.T) {
...
+			input: `"a;b;c;d`,
+			want:  []string{`"a;`, `b;`, `c;`, `d`},
+		},
+		{
+			input: `"a;b";"c;d`,
+			want:  []string{`"a;b";`, `"c;`, `d`},
+		},
+		{
+			// Quotes must be escaped via RFC2047 encoding, not just a backslash.
+			// b through c below must not be treated as a single quoted-run.
```
jhillyerd, line 127: ditto

jhillyerd (changes requested):

internal/stringutil/split_test.go
dmytrokasianenko-outreach: yeah, I was not sure about its original behavior, because it is possible to have `Content-Type: application/rtf; charset=utf-8; name="\"žába\".jpg"` (in [this test](https://github.com/jhillyerd/enmime/blob/49069813358fa4de47947106637d5eecdbe3ef8f/mediatype/mediatype_test.go#L356)), which means quotes can be escaped

internal/stringutil/find.go
dmytrokasianenko-outreach: That's true, it sounds like it would find all quoted `rune`s.

The idea was to have the same naming convention. Before this PR, there was a function `SplitQuoted`, it works like `strings.Split`, but respects quotes. That's why I named this one `FindQuoted`, it finds `rune`, but respects quotes.

I can rename all functions to `FindUnquoted`, `SplitUnquoted` and `SplitAfterUnqioted`, WDYT?

internal/stringutil/split_test.go
dmytrokasianenko-outreach: thanks, comment removed

internal/stringutil/split_test.go
dmytrokasianenko-outreach: thanks, comment removed

internal/stringutil/find.go
jhillyerd: I think renaming everything to Unquoted would make sense

dmytrokasianenko-outreach: fixed 🙂

jhillyerd (approved):

PR #257, opened 2022-06-28
feat: add option to treat multipart w/o boundary as single-part

dmytrokasianenko-outreach:
Thit PR is a proposal to add option to treat multipart message without boundary as single-part.

For example message:
```
Content-Type: multipart/report;

I'm  multipart message without boundary
```

doesn't have boundary parameter in `Content-Type` header, but still can be parsed as single-part message.
Currently, parser returns error that it can't find boundary. With option turned on, parser will treat multipart messages without boundary as single-part.

### DIFF
--- a/detect.go
+++ b/detect.go
@@ -8,14 +8,21 @@ import (
 )
 
 // detectMultipartMessage returns true if the message has a recognized multipart Content-Type header
-func detectMultipartMessage(root *Part) bool {
+func detectMultipartMessage(root *Part, multipartWOBoundaryAsSinglepart bool) bool {
 	// Parse top-level multipart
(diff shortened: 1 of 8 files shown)


jhillyerd:
Overall looks good, only request is to change `Singlepart` to `SinglePart`, as the correct prefixed-word would be `Unipart` but I don't think that would make sense to users.

dmytrokasianenko-outreach:
> Overall looks good, only request is to change `Singlepart` to `SinglePart`, as the correct prefixed-word would be `Unipart` but I don't think that would make sense to users.

updated 🙂

jhillyerd:
Thanks!

Issue #260, opened 2022-08-14
Field accessors for MailBuilder


andylibrian:
Hi, thanks for the project. It's really helpful.

## What I want to do:

I want to access MailBuilder fields from other functions. Apparently the fields are private and there is no accessor available.

```go
func SendEmail(mail *enmime.MailBuilder) error {
        bcc := mail.GetBcc()
}
```

## What I expected:

I would like to access the fields before / without building the MIME part.

## What I got:

No function

## Release or branch I am using:

master

## Note

If you agree to support this, I'm interested to submit a PR to implement it

jhillyerd:
Sure, I'd accept a PR to add getter funcs.

PR #261, opened 2022-08-17
feat: add field accessors for mailbuilder

andylibrian:
Hi, this PR is to implement the field accessors for `MailBuilder` as proposed in #260

### DIFF
--- a/builder.go
+++ b/builder.go
@@ -88,9 +88,11 @@ func (p MailBuilder) ToAddrs(to []mail.Address) MailBuilder {
 	return p
 }
 
-// GetTo returns the stored to addresses.
+// GetTo returns a copy of the stored to addresses.
 func (p *MailBuilder) GetTo() []mail.Address {
-	return p.to
+	var to []mail.Address
+	to = append(to, p.to...)
+	return to
 }
 
 // CC returns a copy of MailBuilder with this name & address appended to the CC header.  name may be
@@ -108,9 +110,11 @@ func (p MailBuilder) CCAddrs(cc []mail.Address) MailBuilder {
 	return p
 }
 
-// GetCC returns the stored cc addresses.
+// GetCC returns a copy of the stored cc addresses.
 func (p *MailBuilder) GetCC() []mail.Address {
-	return p.cc
+	var cc []mail.Address
+	cc = append(cc, p.cc...)
+	return cc
 }
 
 // BCC returns a copy of MailBuilder with this name & address appended to the BCC list.  name may be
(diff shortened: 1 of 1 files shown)


jhillyerd:
I need to take a deeper look, but my initial concern with the values being returned is that they may allow the original value to be mutated -- which goes against the expectations of a Go builder.  I think we'll probably need to do some copying or deep copying on the returned values.

andylibrian:
Hi @jhillyerd, thanks for letting me know your concern. I'm ok with returning the copied values. Once you are sure this is the way to go, please let me know, I'll update the PR to use returned values instead.

builder.go
```diff
@@ -73,6 +88,11 @@ func (p MailBuilder) ToAddrs(to []mail.Address) MailBuilder {
 	return p
 }
 
+// GetTo returns the stored to addresses.
+func (p *MailBuilder) GetTo() []mail.Address {
+	return p.to
```
jhillyerd, line 93: Let's use the following style to return a copy of the to slice:

var dst []string
dst = append(dst, src...)

builder.go
```diff
@@ -88,6 +108,11 @@ func (p MailBuilder) CCAddrs(cc []mail.Address) MailBuilder {
 	return p
 }
 
+// GetCC returns the stored cc addresses.
+func (p *MailBuilder) GetCC() []mail.Address {
+	return p.cc
```
jhillyerd, line 113: ditto for this one

builder.go
```diff
@@ -106,13 +131,23 @@ func (p MailBuilder) BCCAddrs(bcc []mail.Address) MailBuilder {
 	return p
 }
 
+// GetBCC returns the stored bcc addresses.
+func (p *MailBuilder) GetBCC() []mail.Address {
+	return p.bcc
```
jhillyerd, line 136: and this one

builder.go
```diff
@@ -125,18 +160,33 @@ func (p MailBuilder) Header(name, value string) MailBuilder {
...
+
 // Text returns a copy of MailBuilder that will use the provided bytes for its text/plain Part.
 func (p MailBuilder) Text(body []byte) MailBuilder {
 	p.text = body
 	return p
 }
 
+// GetText returns the stored text/plain part.
+func (p *MailBuilder) GetText() []byte {
+	return p.text
```
jhillyerd, line 176: this will also need to be copied.

builder.go
```diff
@@ -125,18 +160,33 @@ func (p MailBuilder) Header(name, value string) MailBuilder {
...
+
 // HTML returns a copy of MailBuilder that will use the provided bytes for its text/html Part.
 func (p MailBuilder) HTML(body []byte) MailBuilder {
 	p.html = body
 	return p
 }
 
+// GetHTML returns the stored text/html part.
+func (p *MailBuilder) GetHTML() []byte {
+	return p.html
```
jhillyerd, line 187: and this one.

jhillyerd (changes requested):
Thanks, looks like we only need to worry about mutability with the slice types.

jhillyerd (approved):

jhillyerd:
Thanks!

PR #262, opened 2022-09-05
feat: ability to recover partial content from broken encoding

fcisnerosrojo:
### --> Problem

**What I did:**
I wrote a simple email which has two parts (text/plain and text/html) with base64 encoded data, but with one characteristic: both part contents have one extra base64 character (so they are slightly broken). Instead of having 4\*n characters, the contents are (4\*n)+1 characters long.

**What I expected:**
ReadParts() returns the root Part, with a FirstChild, and this one with a NextSibling. FirstChild and NextSibling have the partially decoded base64 data (Content != nil, parser recovers partial data from the parts with broken encoding).

**What I got:**
ReadParts() returns the root Part, with a FirstChild, and this one with a NextSibling. FirstChild and NextSibling have Content == nil (Parser does not recover partial data)
The reason is that for each part a `base64.CorruptInputError` is thrown, which then is catched by `base64CorruptInputCheck()`, leaving the part content empty and returning nil error (`skipMalformedParts` flag on true/false would not change the output).

### --> Solution

With this particular scenario, in which there are text/plain and text/html contents, I would like to ignore that specific error (base64.CorruptInputError) and to keep the buffer read at that moment.
But the thing is that there might be different preferences (for now on, "policies"). For example:
- keep the buffer read only when base64.CorruptInputError is raised
- keep the buffer read only when another error is raised
- keep the buffer read only on a base64.CorruptInputError error and when ContentType is text/plain
- keep the buffer read only on a base64.CorruptInputError error and when ContentType is text/plain or text/html
- always keep the buffer read, no matter the error
- and so on ...

To solve this, I added a callback function `readPartErrorPolicy` to the `Parser` type, from which it will be possible to decide when to keep the buffer or not when an error is thrown while reading a part content.

> When no policy is sent to the parser, default behavior is kept.

**Release or branch I am using:**
`master`

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -1,10 +1,14 @@
 package enmime
 
+// ReadPartErrorPolicy allows to recover the buffer (or not) on an error when reading a Part content.
+type ReadPartErrorPolicy func(*Part, error) bool
+
 // Parser parses MIME.
 // Default parser is a valid one.
 type Parser struct {
 	skipMalformedParts              bool
 	multipartWOBoundaryAsSinglePart bool
+	readPartErrorPolicy             ReadPartErrorPolicy
 }
 
 // defaultParser is a Parser with default configuration.
--- a/parser.go
+++ b/parser.go
@@ -1,7 +1,17 @@
 package enmime
 
 // ReadPartErrorPolicy allows to recover the buffer (or not) on an error when reading a Part content.
+//
+// See AllowCorruptTextPartErrorPolicy for usage.
 type ReadPartErrorPolicy func(*Part, error) bool
+
+// AllowCorruptTextPartErrorPolicy recovers partial content from base64.CorruptInputError when content type is text/plain or text/html.
+func AllowCorruptTextPartErrorPolicy(p *Part, err error) bool {
+	if IsBase64CorruptInputError(err) && (p.ContentType == ctTextHTML || p.ContentType == ctTextPlain) {
+		return true
+	}
+	return false
+}
 
 // Parser parses MIME.
 // Default parser is a valid one.
--- a/part.go
+++ b/part.go
@@ -164,7 +164,7 @@ func (p *Part) readPartContent(r io.Reader, readPartErrorPolicy ReadPartErrorPol
 	buf, err := ioutil.ReadAll(r)
 	if err != nil {
 		if readPartErrorPolicy != nil && readPartErrorPolicy(p, err) {
-			p.addWarning(ErrorMalformedChildPart, "keeping the buffer read on error: %s", err.Error())
+			p.addWarning(ErrorMalformedChildPart, "partial content: %s", err.Error())
 			return buf, nil
 		}
 		return nil, err
@@ -314,7 +314,9 @@ func (p *Part) decodeContent(r io.Reader, readPartErrorPolicy ReadPartErrorPolic
 	return nil
 }
 
-// IsBase64CorruptInputError returns true when err is of type base64.CorruptInputError
+// IsBase64CorruptInputError returns true when err is of type base64.CorruptInputError.
+//
+// It can be used to create ReadPartErrorPolicy functions.
 func IsBase64CorruptInputError(err error) bool {
 	switch errors.Cause(err).(type) {
 	case base64.CorruptInputError:
--- a/part.go
+++ b/part.go
@@ -160,11 +160,23 @@ func (p *Part) setupContentHeaders(mediaParams map[string]string) {
 	}
 }
 
+func (p *Part) readPartContent(r io.Reader, readPartErrorPolicy ReadPartErrorPolicy) ([]byte, error) {
+	buf, err := ioutil.ReadAll(r)
+	if err != nil {
+		if readPartErrorPolicy != nil && readPartErrorPolicy(p, err) {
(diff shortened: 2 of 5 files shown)


part.go
```diff
@@ -160,11 +160,23 @@ func (p *Part) setupContentHeaders(mediaParams map[string]string) {
 	}
 }
 
+func (p *Part) readPartContent(r io.Reader, readPartErrorPolicy ReadPartErrorPolicy) ([]byte, error) {
+	buf, err := ioutil.ReadAll(r)
+	if err != nil {
+		if readPartErrorPolicy != nil && readPartErrorPolicy(p, err) {
+			p.addWarning(ErrorMalformedChildPart, "keeping the buffer read on error: %s", err.Error())
```
jhillyerd, line 167: Text of this error doesn't make a lot of sense, could we perhaps call it a partial buffer/content/result/etc?

part_test.go
```diff
@@ -427,6 +427,184 @@ func TestMultiSkipMalformedPart(t *testing.T) {
...
+	}
+	test3 = testData{
+		policy:                       examplePolicy3,
+		expectedPlainTextPartContent: plainContent,
+		expectedErrorPlainTextPart:   keepingBufferWarning,
+		expectedHTMLPartContent:      htmlContent,
+		expectedErrorHTMLPart:        keepingBufferWarning,
+	}
+
+	tests := map[string]testData{
```
jhillyerd, line 515: You can combine map+struct literals to do this in one step that keeps the definition close to the name:
(from https://dave.cheney.net/2019/05/07/prefer-table-driven-tests)

```
func TestSplit(t *testing.T) {
    tests := map[string]struct {
        input string
        sep   string
        want  []string
    }{ 
        "simple":       {input: "a/b/c", sep: "/", want: []string{"a", "b", "c"}}, 
        "wrong sep":    {input: "a/b/c", sep: ",", want: []string{"a/b/c"}},
        "no sep":       {input: "abc", sep: "/", want: []string{"abc"}},
        "trailing sep": {input: "a/b/c/", sep: "/", want: []string{"a", "b", "c"}},
    }

    for name, tc := range tests {
        got := Split(tc.input, tc.sep)
        if !reflect.DeepEqual(tc.want, got) {
            t.Fatalf("%s: expected: %v, got: %v", name, tc.want, got)
        }
    }
}
```

part.go
```diff
@@ -302,18 +314,26 @@ func (p *Part) decodeContent(r io.Reader) error {
 	return nil
 }
 
+// IsBase64CorruptInputError returns true when err is of type base64.CorruptInputError
```
jhillyerd, line 317: Please add note that this is meant for use with ReadPartErrorPolicy

part_test.go
```diff
@@ -427,6 +427,184 @@ func TestMultiSkipMalformedPart(t *testing.T) {
...
+
+	// example policy 3: always recover the buffer read, no matter the error
+	examplePolicy3 := enmime.ReadPartErrorPolicy(func(p *enmime.Part, err error) bool {
+		return true
+	})
+
+	plainContent := "Hello World!"
+	htmlContent := "<p>Hello World</p>"
+
+	ilegalBase64Error := enmime.Error{
```
jhillyerd, line 474:
```suggestion
	illegalBase64Error := enmime.Error{
```

parser.go
```diff
@@ -1,10 +1,14 @@
 package enmime
 
+// ReadPartErrorPolicy allows to recover the buffer (or not) on an error when reading a Part content.
```
jhillyerd, line 3: Ideally we'd have some sort of doc-test example that shows how to use this, but I know that's a significant amount of work.  Could we at least point the user at the unit tests for examples?  Otherwise it's really not clear how to use this.

jhillyerd (changes requested):
Thanks, code looks good overall, but I have some suggestions.

part.go
fcisnerosrojo: Do you mean to create a new error const? Something like this:
```go
ErrorPartialContent := "Partial content" // this goes in error.go
p.addWarning("Partial content", "keeping the buffer read on error: %s", err.Error())
```

Or to just change the detail parameter?
```go
p.addWarning(ErrorMalformedChildPart, "partial content returned: %s", err.Error())
```

jhillyerd: Just the detail parameter text.  We do have an ErrorMalformedBase64 const already, but I'm assuming this error policy could apply to more scenarios than just base64?

parser.go
jhillyerd: I think that would be helpful.  parser.go is pretty short, so I'd be OK with putting it there.

If it's likely we add more of these in the future, a new package might be nice to give them their own Go doc section, and make it easier to copy/paste into your ones own code for modification.

part.go
fcisnerosrojo: Exactly. I'll change then the detail parameter name. Thanks!

fcisnerosrojo:
> Thanks, code looks good overall, but I have some suggestions.

Thanks for your feedback @jhillyerd ! All suggestions were included in the last commit .

jhillyerd:
Thanks!

fcisnerosrojo:
@jhillyerd , can we get a release that includes this feature?

jhillyerd:
done

PR #265, opened 2022-09-18
Update golangci-lint to 1.51.0

jhillyerd:
Blocked until https://github.com/golangci/golangci-lint-action/issues/535 is fixed


jhillyerd:
Given the lack of movement on the reference bug, and work on supporting Go 1.20 in enmime, we can probably migrate our formatting to the new standard and then update the linter version.

Issue #267, opened 2022-11-14
Delivery Status Notification parsing (bounce messages)


dmytrokasianenko-outreach:
Hi James, we have code for parsing of Delivery Status Notification (DSN) per [rfc3464](https://datatracker.ietf.org/doc/html/rfc3464) and [rfc6522](https://datatracker.ietf.org/doc/html/rfc6522) (also known as [bounce message](https://en.wikipedia.org/wiki/Bounce_message)). I can add this parsing logic here. Do you think this kind of logic belongs to enmime or enmime would benefit from it?

jhillyerd:
Yes, that does sound like a useful addition!

PR #268, opened 2022-11-16
feat: parse DSN

dmytrokasianenko-outreach:
This PR adds parsing of Delivery Status Notification (DSN) per [rfc3464](https://datatracker.ietf.org/doc/html/rfc3464) and [rfc6522](https://datatracker.ietf.org/doc/html/rfc6522) (also known as [bounce message](https://en.wikipedia.org/wiki/Bounce_message)). 

Closes https://github.com/jhillyerd/enmime/issues/267

### DIFF
--- a/dsn/dsn.go
+++ b/dsn/dsn.go
@@ -1,0 +1,41 @@
+// Package dsn meant to work with Delivery Status Notification (DSN) per rfc3464:
+// https://datatracker.ietf.org/doc/html/rfc3464
+package dsn
+
+import (
+	"net/textproto"
+	"strings"
+)
+
+// Report represents delivery status report as per https://datatracker.ietf.org/doc/html/rfc6522.
+// It is container of the MIME type multipart/report and contains 3 parts.
+type Report struct {
+	// Explanation contains a human-readable description of the condition(s) that caused the report to be generated.
+	Explanation Explanation
+	// DeliveryStatus provides a machine-readable description of the condition(s) that caused the report to be generated.
+	DeliveryStatus DeliveryStatus
+	// OriginalMessage is optional original message or its portion.
+	OriginalMessage []byte
+}
+
+// Explanation contains a human-readable description of the condition(s) that caused the report to be generated.
+// Where a description of the error is desired in several languages or several media, a multipart/alternative construct MAY be used.
+type Explanation struct {
+	// Text is a text/plain part of the explanation.
+	Text string
+	// HTML is a text/html part of the explanation.
+	HTML string
+}
+
+// DeliveryStatus provides a machine-readable description of the condition(s) that caused the report to be generated.
+type DeliveryStatus struct {
+	// RecipientDSN is Delivery Status Notification per message.
+	MessageDSNs []textproto.MIMEHeader
+	// RecipientDSN is Delivery Status Notification per recipient.
+	RecipientDSNs []textproto.MIMEHeader
+}
+
+// IsFailed returns true if Action field is "failed", meaning message could not be delivered to the recipient.
+func IsFailed(n textproto.MIMEHeader) bool {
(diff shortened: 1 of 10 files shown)


dsn/dsn.go
```diff
@@ -0,0 +1,39 @@
...
+
+// Report represents delivery status report as per https://datatracker.ietf.org/doc/html/rfc6522.
+// It is container of the MIME type multipart/report and contains 3 parts.
+type Report struct {
+	// Explanation contains a human-readable description of the condition(s) that caused the report to be generated.
+	Explanation Explanation
+	// DeliveryStatus provides a machine-readable description of the condition(s) that caused the report to be generated.
+	DeliveryStatus DeliveryStatus
+	// OriginalMessage is optional original message or its portion.
+	OriginalMessage []byte
```
dmytrokasianenko-outreach, line 18: should we have here `[]byte` or `Envelope`?

dsn/testdata/delayed_dsn.raw
```diff
@@ -0,0 +1,27 @@
+MIME-Version: 1.0
```
jhillyerd, line 1: Given we are introducing a number of test data files, I think making a `testdata/dsn` subdirectory would be a good idea.

dsn/dsn.go
jhillyerd: I think it's good to leave as bytes for now, not worth paying CPU cycles for a parse unless the user needs it.  Plus with the new decoding options, we'd probably want them to choose how to parse it.

jhillyerd (changes requested):
Looks good, just a several nit picks

dsn/parse.go
```diff
@@ -0,0 +1,156 @@
...
+			return setExplanation(e, p.FirstChild.NextSibling)
+		}
+	default:
+		return false
+	}
+
+	return false
+}
+
+func parseDeliveryStatus(date []byte) (DeliveryStatus, error) {
```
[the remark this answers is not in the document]
dmytrokasianenko-outreach: thanks, fixed

dsn/parse.go
```diff
@@ -0,0 +1,156 @@
...
+	fields, err := parseDeliveryStatusMessage(date)
+	if err != nil {
+		return DeliveryStatus{}, fmt.Errorf("parse delivey status: %w", err)
```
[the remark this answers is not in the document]
dmytrokasianenko-outreach: thanks, fixed

dsn/parse.go
```diff
@@ -0,0 +1,156 @@
...
+		MessageDSNs:   perMessage,
+		RecipientDSNs: perRecipient,
+	}, nil
+}
+
+// parseDeliveryStatusMessage parses delivery-status message per https://www.rfc-editor.org/rfc/rfc3464#section-2.1
+// The body of a message/delivery-status consists of one or more "fields" formatted according to the ABNF of
+// RFC 822 header "fields". In other words, body of delivery status is multiple headers separated by blank line.
+// First part is per-message fields, following by per-recipient fields.
+func parseDeliveryStatusMessage(data []byte) ([]textproto.MIMEHeader, error) {
```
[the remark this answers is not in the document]
dmytrokasianenko-outreach: agree, renamed

dsn/testdata/delayed_dsn.raw
dmytrokasianenko-outreach: do you want them in `enmine/testdata/dsn/` or in `enmime/dsn/testdata/dsn/`?

jhillyerd: Nevermind on this one, I didn't realize these were already in the `dsn` pkg directory, should be fine as is.

jhillyerd:
Thanks!

PR #270, opened 2023-01-27
fix: Reject invalid MIME header name characters to support Go 1.20

### LINKED ISSUE #269: TestReadHeader/equals_in_name fails on Go1.20
What I did:

```
$ go version
go version go1.20rc3 linux/amd64
```

```
go test -run TestReadHeader/equals_in_name
```

What I expected: PASS

What I got:

```
--- FAIL: TestReadHeader (0.00s)
    --- FAIL: TestReadHeader/equals_in_name (0.00s)
        header_test.go:325: malformed MIME header line: name=value:text
FAIL
```

Release or branch I am using: master

jhillyerd:
Fixes #269

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -8,15 +8,15 @@ jobs:
       matrix:
         go:
           - 1.16.x
-          - 1.17.x
           - 1.18.x
+          - 1.20.0-rc.2
     name: Go ${{ matrix.go }} build
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
         with:
           fetch-depth: 0
       - name: Setup Go
-        uses: actions/setup-go@v2
+        uses: actions/setup-go@v3
         with:
           go-version: ${{ matrix.go }}
       - name: Build and Test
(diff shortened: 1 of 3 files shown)


kayabe:
test fails because of this commit https://github.com/golang/go/commit/a6642e67e16b9d769a0c08e486ba08408064df19

jhillyerd:
Yes, as seems to be the usual with enmime, we have to decide how to handle malformed email in an elegant way.

It does not appear there is an official way to escape characters in mime header names.   The easiest would probably be to convert the equals to a dash.  I could imagine that being exploited, so perhaps a double dash would be better.  Might also provide a strict mode to discard that particular header line instead.

rjbordon:
Hi @jhillyerd, first of all, thanks for this awesome library!

I have been testing the latest and greatest in the master branch (this PR included) with go1.20.1 and it works out as expected. Could you kindly create a new release from the master branch? I think the last one goes until Sept 2022.

Thanks.

jhillyerd:
done, v0.11.0 tagged

rjbordon:
Thank you!

PR #284, opened 2023-03-07
Read all part content use ReadPartErrorPolicy.

ly020044:
Hi,
    
The `ReadPartErrorPolicy` only used by content type `text/*`,  Wouldn't it be better to use ReadPartErrorPolicy to control known errors when reading all part content?

For example, if the content type is `image/*`, a `base64.CorruptInputError` occurs while reading part content. This image may also be displayed if this error is ignored.

### DIFF
--- a/part.go
+++ b/part.go
@@ -301,7 +301,7 @@ func (p *Part) decodeContent(r io.Reader, readPartErrorPolicy ReadPartErrorPolic
 		}
 	}
 	// Decode and store content.
-	content, err := ioutil.ReadAll(contentReader)
+	content, err := p.readPartContent(contentReader, readPartErrorPolicy)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


iredmail:
This PR fixes errors like `[E] Malformed Base64: illegal base64 data at input byte 0`.

We already processed many emails with this error, the cause is base64 encoded attachment (e.g. image) is incomplete, which causes parse error like above one.

If we open this email with MUA like Thunderbird or macOS Mail.app, email body is fine, but the image is partly rendered (but yes we can still view part of the image).

With this PR, we can define a policy like `AllowCorruptTextPartErrorPolicy` (but for all content types) to ignore base64 corrupt input, then successfully parse the email and get the incomplete image (which is expected).

PR #288, opened 2023-04-15
feat: add support for multiple Reply-To addresses

nolotz:
Hello @jhillyerd and enmime maintainers,

I've been working on a feature enhancement for the enmime library to add support for multiple reply-to addresses, as described in RFC 5322, section 3.6.2. I believe this functionality will improve the flexibility of the library and provide users with the ability to correctly handle multiple reply-to addresses as per the email specification.

Here is a brief overview of the changes I've made in this commit:

- Updated the ``MailBuilder`` struct to store multiple reply-to addresses.
- Modified the ``Build`` function to correctly write multiple reply-to addresses.
- Updated the ``GetReplyTo`` function to return the first email in the list for backward compatibility, and added a new ``GetReplyToAddrs`` function to return a slice copy of all addresses.
- Updated the existing ``ReplyTo`` function to support backward compatibility by handling single email addresses while using the new reply-to address list, and added a new ``ReplyToAddrs`` function to support setting the whole reply-to list.
- Added new tests to validate the proper handling of multiple reply-to addresses.

Thank you for your time and consideration. I'm looking forward to your feedback and the opportunity to contribute to this project.

### DIFF
--- a/builder.go
+++ b/builder.go
@@ -22,7 +22,7 @@ import (
 type MailBuilder struct {
 	to, cc, bcc          []mail.Address
 	from                 mail.Address
-	replyTo              mail.Address
+	replyTo              []mail.Address
 	subject              string
 	date                 time.Time
 	header               textproto.MIMEHeader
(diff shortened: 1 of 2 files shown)


builder.go
```diff
@@ -153,13 +153,33 @@ func (p *MailBuilder) GetBCC() []mail.Address {
...
+// ReplyToAddrs returns a copy of MailBuilder with the new reply to header list. This method only
+// has an effect if the Send method is used to transmit the message, there is no effect on the parts
+// returned by Build().
+func (p MailBuilder) ReplyToAddrs(replyTo []mail.Address) MailBuilder {
+	p.replyTo = replyTo
 	return p
 }
 
-// GetReplyTo returns the stored replyTo header.
+// GetReplyTo returns the first stored replyTo header.
```
jhillyerd, line 170: Given enmime is pre-1.0, I'd prefer changing the API contract here and just having GetReplyTo return the slice instead so that it matches GetTo & GetCC

jhillyerd (changes requested):
Thanks, good idea

builder.go
iredmail: Hi @jhillyerd 

If you're going to break API contract in next release, do you accept PR to keep original header cases (lower / upper)?
https://github.com/jhillyerd/enmime/pull/283#issuecomment-1464835088

jhillyerd: @iredmail yeah, might as well do that at the same time.

jhillyerd:
Thanks!

PR #289, opened 2023-04-27
fix for values with encoded equal signs

komkom:
We faced the issue that emails with encoded equal signs could not be parsed in certain cases. This fix should improve the situation. 
Thank you in advance for any code reviews and great library!


iredmail:
Better use `strings.Cut()` instead of `strings.SplitAfterN()`.

- `strings.Cut()` was introduced in Go 1.18.
- enmime requires Go 1.18 too as defined in `go.mod`.

PR #296, opened 2023-07-23
add io.Reader interface for attachment content

erankor:
can provide a significant optimization in memory usage when encoding large attachments - the attachment is stored only once (in the final message), instead of 3 times (part.Content, the base64 encoded buffer and the final message)

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -26,12 +26,28 @@ const (
 	teBase64
 )
 
+const (
+	base64EncodedLineLen = 76
+	base64DecodedLineLen = base64EncodedLineLen * 3 / 4 // this is ok since lineLen is divisible by 4
+	linesPerChunk        = 128
+	readChunkSize        = base64DecodedLineLen * linesPerChunk
+)
+
 var crnl = []byte{'\r', '\n'}
 
 // Encode writes this Part and all its children to the specified writer in MIME format.
 func (p *Part) Encode(writer io.Writer) error {
 	if p.Header == nil {
 		p.Header = make(textproto.MIMEHeader)
+	}
+	if p.ContentReader != nil {
+		// read some data in order to check whether the content is empty
+		p.Content = make([]byte, readChunkSize)
+		n, err := p.ContentReader.Read(p.Content)
+		if err != nil && err != io.EOF {
+			return err
+		}
+		p.Content = p.Content[:n]
 	}
 	cte := p.setupMIMEHeaders()
 	// Encode this part.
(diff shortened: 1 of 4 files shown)


encode_test.go
```diff
@@ -104,6 +106,78 @@ func TestEncodePartQuotedPrintableHeaders(t *testing.T) {
 	test.DiffGolden(t, b.Bytes(), "testdata", "encode", "part-quoted-printable-headers.golden")
 }
 
+type oneByOneReader struct {
```
jhillyerd, line 109: Please add comment on the purpose of this to make tests easier to understand

encode_test.go
```diff
@@ -104,6 +106,78 @@ func TestEncodePartQuotedPrintableHeaders(t *testing.T) {
...
+			p.FileName = "stuff.zip"
+			p.FileModDate, _ = time.Parse(time.RFC822, "01 Feb 03 04:05 GMT")
+
+			p.Content = make([]byte, contentLength)
+			_, err := rand.Read(p.Content)
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			// encode the part using byte slice
```
jhillyerd, line 152:
```suggestion
			// encode the part using `Content` byte slice stored in the Part
```

jhillyerd (changes requested):
Thanks for this.  One concern I have is that the builder was designed to be safe to "fork" in the sense that it always returns a copy of itself.  ie you could setup the content, store that copy, then loop over your recipients, adding each one and sending the returned copy.

I suspect storing a content reader would mean only the first use of any particular set of copies would function, and the rest would just see an EOF.  Not the end of the world, but I think it needs to be called out in the builder doc comment.

encode.go
```diff
@@ -205,6 +225,54 @@ func (p *Part) encodeContent(b *bufio.Writer, cte transferEncoding) (err error)
 	return err
 }
 
+// encodeContentFromReader writes out the content read from the reader using base64 encoding.
+func (p *Part) encodeContentFromReader(b *bufio.Writer) error {
+	text := make([]byte, base64EncodedLineLen) // a single base64 encoded line
+	enc := base64.StdEncoding
+
+	chunk := make([]byte, readChunkSize) // contains a whole number of lines
+	copy(chunk, p.Content)
```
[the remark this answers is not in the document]
erankor: done

encode_test.go
erankor: done

encode_test.go
erankor: done

erankor:
Thanks @jhillyerd! added the requested comments

erankor:
ping :)

jhillyerd:
Looks good, thanks!

PR #297, opened 2023-08-08
MIMEHeader: use Go's net/textproto for public interface

### LINKED ISSUE #294: Type of the `(*enmime.Part).Header` field is now private
What I did:
I've had a function that returns the `textproto.MIMEHeader` value from the `(*enmime.Part).Header` field.

What I expected:
That I would continue to be able to return that field.

What I got:
After upgrading, the type of that field is now private, so I cannot return it. Instead, I have to manually convert it from the `enimime` internal `MIMEHeader` into the `textproto.MIMEHeader` from the stdlib.

Release or branch I am using:
v1.0.0

This is a result of #283.

jhillyerd:
We did not intend for the external API to change when hosting our own version
of textproto.

Fixes #294

### DIFF
--- a/envelope.go
+++ b/envelope.go
@@ -5,13 +5,13 @@ import (
 	"io"
 	"mime"
 	"net/mail"
-	nettp "net/textproto"
+	"net/textproto"
 	"strings"
 	"time"
 
 	"github.com/jaytaylor/html2text"
 	"github.com/jhillyerd/enmime/internal/coding"
-	"github.com/jhillyerd/enmime/internal/textproto"
+	inttp "github.com/jhillyerd/enmime/internal/textproto"
 	"github.com/jhillyerd/enmime/mediatype"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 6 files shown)


envelope.go
```diff
@@ -26,8 +27,8 @@ type Envelope struct {
 	// All non-text parts that were not placed in Attachments or Inlines, such as multipart/related
 	// content.
 	OtherParts []*Part
-	Errors     []*Error              // Errors encountered while parsing
-	header     *textproto.MIMEHeader // Header from original message
+	Errors     []*Error          // Errors encountered while parsing
+	header     *nettp.MIMEHeader // Header from original message
```
dcormier, line 31: I think this would be cleaner for people looking at docs if you rename the import of `.../internal/textproto` instead of `net/textproto`.

part.go
```diff
@@ -28,11 +29,11 @@ const (
...
-	PartID      string               // PartID labels this part's position within the tree.
-	Parent      *Part                // Parent of this part (can be nil.)
-	FirstChild  *Part                // FirstChild is the top most child of this part.
-	NextSibling *Part                // NextSibling of this part.
-	Header      textproto.MIMEHeader // Header for this part.
+	PartID      string           // PartID labels this part's position within the tree.
+	Parent      *Part            // Parent of this part (can be nil.)
+	FirstChild  *Part            // FirstChild is the top most child of this part.
+	NextSibling *Part            // NextSibling of this part.
+	Header      nettp.MIMEHeader // Header for this part.
```
dcormier, line 36: I think this would be cleaner for people looking at docs if you rename the import of `.../internal/textproto` instead of `net/textproto`.

PR #299, opened 2023-08-14
Rollback pkg/errors portion of #298

jhillyerd:
PR #298 removed usage of the archived package github.com/pkg/errors, and
references to the deprecated ioutil package.

After understanding that stdlib errors still does not support stack trace
wrapping (see discussion in #298), we have decided to rollback these changes
as parser debugging heavily benefits from stack traces.

The ioutil changes are desirable and will be kept.

See also #94

### DIFF
--- a/boundary_test.go
+++ b/boundary_test.go
@@ -3,10 +3,11 @@ package enmime
 import (
 	"bufio"
 	"bytes"
-	"errors"
 	"io"
 	"strings"
 	"testing"
+
+	"github.com/pkg/errors"
 )
 
 func TestBoundaryReader(t *testing.T) {
@@ -462,7 +463,7 @@ func TestBoundaryReaderReadErrors(t *testing.T) {
 	if n != 0 {
 		t.Fatal("Read() should not have read any bytes, failed")
 	}
-	if !errors.Is(err, bufio.ErrBufferFull) {
+	if errors.Cause(err) != bufio.ErrBufferFull {
 		t.Fatal("Read() should have returned bufio.ErrBufferFull error, failed")
 	}
 	// Next method to return a non io.EOF error.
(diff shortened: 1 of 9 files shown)


boundary_test.go
```diff
@@ -462,15 +463,15 @@ func TestBoundaryReaderReadErrors(t *testing.T) {
...
+	if errors.Cause(err) != bufio.ErrBufferFull {
 		t.Fatal("Read() should have returned bufio.ErrBufferFull error, failed")
 	}
 	// Next method to return a non io.EOF error.
 	next, err := br.Next()
 	if next {
 		t.Fatal("Next() should have returned false, failed")
 	}
-	if !errors.Is(err, bufio.ErrBufferFull) {
+	if errors.Cause(err) != bufio.ErrBufferFull {
```
dcormier, line 474: `errors.Is()` is probably better here, ultimately.

inspect.go
```diff
@@ -39,9 +40,10 @@ func DecodeHeaders(b []byte, addtlHeaders ...string) (textproto.MIMEHeader, erro
 	b = ensureHeaderBoundary(b)
 	tr := textproto.NewReader(bufio.NewReader(bytes.NewReader(b)))
 	headers, err := tr.ReadMIMEHeader()
-
-	// io.EOF is expected
-	if err != nil && !errors.Is(err, io.EOF) {
+	switch errors.Cause(err) {
+	case nil, io.EOF:
+	// carry on, io.EOF is expected
+	default:
```
dcormier, line 46: The previous logic, using `errors.Is()`, is probably better here.

part.go
```diff
@@ -325,11 +325,12 @@ func (p *Part) decodeContent(r io.Reader, readPartErrorPolicy ReadPartErrorPolic
...
-	if err == nil {
+	switch errors.Cause(err).(type) {
+	case base64.CorruptInputError:
+		return true
+	default:
 		return false
 	}
-	_, ok := err.(base64.CorruptInputError)
-	return ok
 }
```
dcormier, line 334:
```golang
func IsBase64CorruptInputError(err error) bool {
	 var errCorrupt base64.CorruptInputError
	 return errors.As(err, &errCorrupt)
}
```

dcormier (commented):
Using `errors.Is()` and `errors.As()` to compare error values and types (respectively), and to get inside wrapped errors, is the way to go. Not critical to do right this moment, but it'll make the code more resilient to future changes.

PR #304, opened 2023-10-09
feat: add parser option to parse raw content

jerjako:
Add an option to the parser to skip the auto-detect of the CTE. And return the content unchanged


jhillyerd:
I'm in favor of this, but have two concerns:

1. I don't like that a "Parser" option is changing the output of the encoder.  Given we don't have a way to set encoding options, I think we can live with this for now.
2. This change must add unit tests, otherwise it will be too easy to break in the future.

jerjako:
Thank you @jhillyerd
I know what you mean, I'm not a fan either. I've tested to change the behavior of the Encode function, but without luck.
I've added a simple test that will show the diff with and without the option.

testdata/encode/parser-raw-content-option-false.raw
jhillyerd, line 1: Let's use the `.golden` extension instead of `.raw` to be consistent with the other golden tests in the project.

testdata/encode/parser-raw-content-option-true.raw
jhillyerd, line 1: Let's use the `.golden` extension instead of `.raw` to be consistent with the other golden tests in the project.

testdata/encode/parser-raw-content-option.raw
jhillyerd, line 1: This can keep `.raw` as it's an input.

jhillyerd (changes requested):
Looks good, just a couple small changes, thanks.

testdata/encode/parser-raw-content-option.raw
jerjako: Ok, done 👍

testdata/encode/parser-raw-content-option-false.raw
jerjako: Oh yes, I miss that, thank you.

jhillyerd:
Thanks!

PR #306, opened 2023-10-11
Export readHeader as ReadHeader.

davrux:
The function is excellent at parsing SMTP and MIME headers. It would be great to use this function in other projects.


iredmail:
Is it necessary to export `AddError` and `AddWarning`?

davrux:
Yes, without export an external package won't be able to implement the interface.

jhillyerd:
Sounds like a good idea to me, thanks!

Issue #293, opened 2023-06-22
Request: Additional case-normalizing before passing into mime.ParseMediaType


CyrusRoshan:
I currently have an email header value of `text/html; charset=utf-8; CHARSET=UTF-8` (in an email from Air Canada, if you were wondering), similar to the `text/html; charset=utf-8; charset=UTF-8` reported [here](https://github.com/golang/go/issues/28618#issuecomment-500185459).

I think lowercasing the `ctype` value in `mediatype.Parse` before passing into `mime.ParseMediaType` would fix this case with the the intended behavior.

PR #308, opened 2023-10-13
add new option SetCustomParseMediaType to customise mediatype parsing

vadzappa:
### Problem

we've encountered real mail bodies where `Content-Type` is set to `"application/Pamir Viewer; name=\"some-thing.pmrv\""`, which is not RFC compliant, but is still valuable for us to process it, so it can't be skipped using `SkipMalformedParts`

### Proposed solution

New option `SetCustomParseMediaType` which accepts function which has signature exactly the same as `ParseMediaType`. This option can be utilized in similar scenarios, see provided example: `ExampleSetCustomParseMediaType`

will allow to fix this https://github.com/jhillyerd/enmime/issues/293 indirectly

### DIFF
--- a/options.go
+++ b/options.go
@@ -64,7 +64,8 @@ func (o rawContentOption) apply(p *Parser) {
 	p.rawContent = bool(o)
 }
 
-// SetCustomParseMediaType if provided, will be used to parse media type instead of default ParseMediaType
+// SetCustomParseMediaType if provided, will be used to parse media type instead of the default ParseMediaType
+// function.  This may be used to parse media type parameters that would otherwise be considered malformed.
 // By default parsing happens using ParseMediaType
 func SetCustomParseMediaType(customParseMediaType CustomParseMediaType) Option {
 	return parseMediaTypeOption(customParseMediaType)
--- a/options.go
+++ b/options.go
@@ -63,3 +63,15 @@ type rawContentOption bool
 func (o rawContentOption) apply(p *Parser) {
 	p.rawContent = bool(o)
 }
+
+// SetCustomParseMediaType if provided, will be used to parse media type instead of default ParseMediaType
+// By default parsing happens using ParseMediaType
+func SetCustomParseMediaType(customParseMediaType CustomParseMediaType) Option {
+	return parseMediaTypeOption(customParseMediaType)
+}
+
+type parseMediaTypeOption CustomParseMediaType
+
+func (o parseMediaTypeOption) apply(p *Parser) {
+	p.customParseMediaType = CustomParseMediaType(o)
+}
--- a/options_test.go
+++ b/options_test.go
@@ -1,0 +1,121 @@
+package enmime
+
+import (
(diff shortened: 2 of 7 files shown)


options.go
```diff
@@ -51,3 +51,15 @@ func (o maxStoredPartErrorsOption) apply(p *Parser) {
 	max := int(o)
 	p.maxStoredPartErrors = &max
 }
+
+// SetCustomParseMediaType if provided, will be used to parse media type instead of default ParseMediaType
```
jhillyerd, line 55:
```suggestion
// SetCustomParseMediaType if provided, will be used to parse media type instead of the default ParseMediaType
// function.  This may be used to parse media type parameters that would otherwise be considered malformed.
```

options_test.go
```diff
@@ -0,0 +1,121 @@
...
+
+func ExampleSetCustomParseMediaType() {
+	// for the sake of simplicity replaces space in a very specific invalid content-type: "application/Pamir Viewer"
+	replaceSpecificContentType := func(ctype string) (mtype string, params map[string]string, invalidParams []string, err error) {
+		modifiedStr := strings.ReplaceAll(ctype, "application/Pamir Viewer", "application/PamirViewer")
+
+		return ParseMediaType(modifiedStr)
+	}
+
+	invalidMessageContent := `Return-path: <enmime.one@parser.git>
```
jhillyerd, line 63: I like that this is inline content for the example function, but I think we should remove any unnecessary headers from the message to keep it compact and readable.

part.go
```diff
@@ -324,6 +322,19 @@ func (p *Part) decodeContent(r io.Reader, readPartErrorPolicy ReadPartErrorPolic
 	return nil
 }
 
+// parses media type using custom or default media type parser
+func (p *Part) parseMediaType(ctype string) (mtype string, params map[string]string, invalidParams []string, err error) {
+	if p.parser == nil {
+		return ParseMediaType(ctype)
+	}
+
+	if p.parser.customParseMediaType == nil {
```
jhillyerd, line 331: Could be combined with if statement above.

jhillyerd (changes requested):

options_test.go
vadzappa: since it's a bit unclear from contribution file, would it be feasible to keep commits for fixing proposals as separate commits, or do a rebase after review, so PR will still one commit?

jhillyerd:
I always squash PR merges via the GitHub so you don't need to squash/rebase yourself.

jhillyerd (approved):

jhillyerd:
Looks good now, but I think we have some merge conflicts after I merged the previous PR

vadzappa:
> Looks good now, but I think we have some merge conflicts after I merged the previous PR

did rebase, should be no conflicts now

jhillyerd (approved):

jhillyerd:
Thanks!

PR #310, opened 2023-11-08
feat: don't auto detect charset when using raw content parser option

jerjako:
When the content is sent as raw, the charset can sometimes return a wrong value.
In this text example with Japanese characters, the charset content returning was iso-8859-1.


encode_test.go
```diff
@@ -376,11 +376,11 @@ func TestParseRawContentOptionTrue(t *testing.T) {
 	if err := e.Root.Encode(b); err != nil {
 		t.Fatal(err)
 	}
-	test.DiffGolden(t, b.Bytes(), "testdata", "encode", "parser-raw-content-option-true.raw.golden")
+	test.DiffGolden(t, b.Bytes(), "testdata", "encode", "parser-raw-content-html-option-true.raw.golden")
```
jhillyerd, line 379: Could you add a comment above each of the four TestParseRawContentXXXXX test.DiffGolden calls stating what the test is verifying/expecting?  It won't be obvious to someone that may have to regenerate the golden files in the future.

jerjako: Hi, no problem. I've added the comments

jhillyerd:
Thank you!

PR #313, opened 2023-12-12
feat: skip empty parameters

milankonir:
Looks like there was never a proper condition for filtering out empty parameters. It kinda worked thanks to this https://github.com/jhillyerd/enmime/blob/main/mediatype/mediatype.go#L125-L128 (`=` substring would always match if there was at least one parameter already added to `mtype`). It did not work if the empty parameter was the first parameter.


jhillyerd:
Thanks!

PR #316, opened 2023-12-27
Return public `net/textproto` error types

dcormier:
To allow for checking returned error types, types from the standard library's `net/textproto` must be used instead of types from the private `textproto` fork.

Kind of related to #294.


jhillyerd:
Looks good, thanks

PR #317, opened 2024-01-03
feat: remove tspecials from type/subtype

milankonir:
According to the RFC, content type and subtype cannot contain special characters and any such character will fail parsing. Removing the characters from the type/subtype can help successfully parsing the content type that contains some extra garbage.


testdata/parts/empty-ctype-bad-content.raw
```diff
@@ -1,7 +1,7 @@
 Content-Type: multipart/alternative; boundary="----=_Part_10541_215446765.98298273965074084"
 
 ------=_Part_10541_215446765.98298273965074084
-Content-Type: <html>
+Content-Type:
```
milankonir, line 4: This had to be adjusted as `<html>` was no longer failing validation due to `<` and `>` being removed in `fixMangledMediaType`

jhillyerd:
Part of the goal of enmime is to flag bad input, I know that we were already silently trying to fix up borderline media type strings; but this character stripping is quite aggressive in my opinion.

I'm OK making enmime more resilient to malformed messages, but they should either be optional or add an [Error to the Part](https://github.com/jhillyerd/enmime/blob/main/part.go#L45).  Adding Errors feels like the more useful solution here, but I'm not sure what effort level is required to implement it.

milankonir:
> Part of the goal of enmime is to flag bad input, I know that we were already silently trying to fix up borderline media type strings; but this character stripping is quite aggressive in my opinion.

My reasoning behind the change was that any special character in the content type/subtype will fail the default mime parser while stripping out any garbage can result in a valid content type, but I perfectly understand you don't want to make this the default behavior.

> I'm OK making enmime more resilient to malformed messages, but they should either be optional or add an [Error to the Part](https://github.com/jhillyerd/enmime/blob/main/part.go#L45). Adding Errors feels like the more useful solution here, but I'm not sure what effort level is required to implement it.

I have made the invalid character stripping opt-in via the parser option. Let me know if this is acceptable.

Happy to address any other issues you see with the PR.

header.go
```diff
@@ -113,10 +113,10 @@ func ParseAddressList(list string) ([]*mail.Address, error) {
 //   - Missing ';' between content-type and media parameters
 //   - Repeating media parameters
 //   - Unquoted values in media parameters containing 'tspecials' characters
```
jhillyerd, line 115:
```suggestion
//   - Unquoted values in media parameters containing 'tspecials' characters
//
// Deprecated: Use mediatype.Parse instead.
```

header.go
```diff
@@ -113,10 +113,10 @@ func ParseAddressList(list string) ([]*mail.Address, error) {
...
-func ParseMediaType(ctype string) (mtype string, params map[string]string, invalidParams []string,
```
jhillyerd, line 116: Instead of changing the signature of this function, let's just have it default to not stripping the invalid characters, and deprecate the entire function instead.  It's redundant now that our `mediatype` is a public package.

mediatype/mediatype.go
```diff
@@ -30,16 +30,20 @@ const (
...
+}
+
 // Parse is a more tolerant implementation of Go's mime.ParseMediaType function.
 //
 // Tolerances accounted for:
 //   - Missing ';' between content-type and media parameters
 //   - Repeating media parameters
 //   - Unquoted values in media parameters containing 'tspecials' characters
 //   - Newline characters
-func Parse(ctype string) (mtype string, params map[string]string, invalidParams []string, err error) {
```
jhillyerd, line 40: Now that enmime is post 1.0, changing the public API is a big deal.  I'd recommend renaming this function to `ParseWithOptions` or similar, and adding a wrapper `Parse` that uses default options to maintain the original API.

jhillyerd (changes requested):

milankonir:
Apologies for the long loop, should be adjusted per your comments.

mediatype/mediatype.go
```diff
@@ -38,8 +42,13 @@ const (
 //   - Unquoted values in media parameters containing 'tspecials' characters
 //   - Newline characters
 func Parse(ctype string) (mtype string, params map[string]string, invalidParams []string, err error) {
+	return ParseWithOptions(ctype, MediaTypeParseOptions{})
+}
+
+// Parse method with additional options controlling the parsing behavior
```
jhillyerd, line 48:
```suggestion
// ParseWithOptions parses media-type with additional options controlling the parsing behavior.
```

our godoc linter is broken right now

jhillyerd (changes requested):
Looks good, couple small changes

jhillyerd (approved):

jhillyerd:
Thank you!

PR #318, opened 2024-01-03
fix: don't encode header value with the RawContent parser option

jerjako:
Hi, with this PR, I want to keep my header values like the original with the raw content parser option set to true.

### DIFF
--- a/encode_test.go
+++ b/encode_test.go
@@ -426,7 +426,7 @@ func TestParseRawContentTextOptionFalse(t *testing.T) {
 	test.DiffGolden(t, b.Bytes(), "testdata", "encode", "parser-raw-content-text-option-false.raw.golden")
 }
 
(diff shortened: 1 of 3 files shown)


encode_test.go
```diff
@@ -425,3 +425,19 @@ func TestParseRawContentTextOptionFalse(t *testing.T) {
 	}
 	test.DiffGolden(t, b.Bytes(), "testdata", "encode", "parser-raw-content-text-option-false.raw.golden")
 }
+
+// TestRawContentUTF8Headers test if not encoded UTF8 headers are not edited with the rawContent parser option.
```
jhillyerd, line 429: Would this be more correct?

```suggestion
// TestRawContentUTF8Headers verifies plain-text headers are unmodified with the rawContent parser option.
```

jhillyerd:
Thanks!

PR #319, opened 2024-01-07
chore: fixed timezone for tests

milankonir:
The tests are failing when ran in a TZ different from GMT.

```
--- FAIL: TestEncodePartDefaultHeaders (0.00s)
    encode_test.go:71: Test output did not match testdata/encode/part-default-headers.golden
        To update golden file, run: go test -update
    encode_test.go:71: diff -want +got:
        | Content-Disposition: attachment; filename=stuff.zip; modification-date="01
        |- Feb 03 04:05 GMT"
        |+ Feb 03 05:05 CET"
        | Content-Id: <mycontentid>
        | Content-Transfer-Encoding: base64
        ...
        | WklQWklQWklQ
        | 
```

This PR fixates test run to TZ=Europe/London

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ build:
 	go build
 
 test:
-	TZ=Europe/London go test -race ./...
+	go test -race ./...
 
 fmt:
 	@gofmt -l -w $(SRC)
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ build:
 	go build
 
 test:
-	go test -race ./...
+	TZ=Europe/London go test -race ./...
 
 fmt:
 	@gofmt -l -w $(SRC)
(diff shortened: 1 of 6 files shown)


jhillyerd:
For the record, I am not in GMT/UTC, and am unable to duplicate the failure with any of:

```sh
env TZ=CET go test -race -count 1 ./...
env TZ=America/Los_Angeles go test -race -count 1 ./...
env TZ=Europe/London go test -race -count 1 ./...
```

Perhaps this is a change in a newer version of Go than the 1.20.11 I am running.  Either way, I don't think altering the Makefile is the correct way to fix this, as IDEs and GitHub CI won't be using Makefiles.

Someone contributed a similar fix to one of my other projects in https://github.com/inbucket/inbucket/pull/406/files -- that's likely to be a better approach, _if_ we can use it here.

milankonir:
So this is quite interesting, the test were failing for my TZ specifically. You can reproduce yourself by running:

```
TZ=Europe/Prague go test -race ./... 
```

I assume it has to do with zone abbreviation vs current locale behavior described in https://pkg.go.dev/time#Parse

Originally, I did not want to make code changes for the sake of tests, but formatting the mod date in UTC should not do any harm.

jhillyerd:
Ah, yes, your example does fail for me, interesting.

Thanks for sending in the fix.

PR #334, opened 2024-05-03
Add parser option to disable HTML to Text down conversion

tvdv:


jhillyerd:
Good idea!  Sorry for the slow response, I wanted to rename the option, but then saw we needed tests as well.

PR #342, opened 2024-08-29
feat: add option to disable character detection

### LINKED ISSUE #340: Decoding issue (Used in Mailpit)
A tool we use at work, uses your library. It's creator forwarded me to your library, so here I am. The original issue is at https://github.com/axllent/mailpit/issues/348

@axllent also commented a go example for smth, you might wanna check it out.

Basically, an email shown by this tool formats the text with wrong characters.

Here are a working case (any length shorter or equal to this will work) 

[example-works.eml.txt](https://github.com/user-attachments/files/16622607/example-works.eml.txt)

image

And an example where anything longer or equal to this length (not sure if this is actually true) will break

[example-broken.eml.txt](https://github.com/user-attachments/files/16622612/example-broken.eml.txt)

image

when being sent to any other mailcatcher software or opening in MS Outlook, it is displayed correctly

image

I hope this can help you identify the issue :)

axllent:
This adds an option to disable character detection (when detected in the message part).

Resolves #340

### DIFF
--- a/options.go
+++ b/options.go
@@ -100,3 +100,15 @@ func (o disableTextConversionOption) apply(p *Parser) {
 func DisableTextConversion(disableTextConversion bool) Option {
 	return disableTextConversionOption(disableTextConversion)
 }
+
+type disableCharacterDetectionOption bool
+
+func (o disableCharacterDetectionOption) apply(p *Parser) {
+	p.disableCharacterDetection = bool(o)
+}
+
+// DisableCharacterDetection sets the disableCharacterDetection option. When true, the parser will use the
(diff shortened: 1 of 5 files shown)


jhillyerd (approved):

jhillyerd:
Thanks!  I also re-ordered things a bit in #343 so we don't waste CPU on charset detection when its disabled.

axllent:
That's awesome, it makes perfect sense to shift that check up to prevent those unnecessary CPU cycles @jhillyerd :+1:  Looking forward to a new release! Thanks.

Issue #358, opened 2024-12-10
Enhancement: content transfer encoding


pgradoboev:
Sometimes, when creating a MIME message, it is necessary to explicitly set the "quoted-printable" transfer encoding for a certain part instead of automatic detection.

I would like to know your opinion (before creating a PR), will it be useful to anyone else?

jhillyerd:
This sounds like a good idea to me -- but I don't do much encoding so I can't guess how valuable to others.

Given the current 2.0.0 release was API breaking, if you do send a PR, please make it additive, otherwise I would not want to merge it for some time.

PR #359, opened 2024-12-13
feat: encoding options added

pgradoboev:
James, please look at the proposed changes. #358

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -117,7 +117,7 @@ func (p *Part) setupMIMEHeaders() transferEncoding {
 		} else {
 			cte = teBase64
 			if p.TextContent() && p.ContentReader == nil {
-				cte = selectTransferEncoding(p.Content, false)
+				cte = p.selectTransferEncoding(p.Content, false)
 				if p.Charset == "" {
 					p.Charset = utf8
 				}
@@ -144,7 +144,7 @@ func (p *Part) setupMIMEHeaders() transferEncoding {
 		p.Header.Set(hnContentID, coding.ToIDHeader(p.ContentID))
 	}
 	fileName := p.FileName
-	switch selectTransferEncoding([]byte(p.FileName), true) {
+	switch p.selectTransferEncoding([]byte(p.FileName), true) {
 	case teBase64:
 		fileName = mime.BEncoding.Encode(utf8, p.FileName)
 	case teQuoted:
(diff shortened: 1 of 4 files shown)


encoder_options.go
```diff
@@ -0,0 +1,34 @@
+package enmime
+
+type EncoderOption interface {
+	apply(p *Encoder)
+}
+
+// Encoder implements MIME part encoding options
+type Encoder struct {
+	// Enforce Quoted encoding when selecting Content Transfer Encoding
+	enforceQuotedCte bool
```
jhillyerd, line 10: In my opinion, `enforce` is the wrong term, as quoted-cte is not a rule or policy.  Just use `force` here and elsewhere.

encoder_options.go
```diff
@@ -0,0 +1,34 @@
...
+}
+
+// EnforceQuotedPrintableCte enforces "quoted-printable" transfer encoding when selecting Content Transfer Encoding.
```
jhillyerd, line 13:
```suggestion
// EnforceQuotedPrintableCte forces "quoted-printable" transfer encoding when selecting Content Transfer Encoding, preventing the use of base64.
```

encode.go
```diff
@@ -303,12 +303,18 @@ func (p *Part) encodeContentFromReader(b *bufio.Writer) error {
 }
 
 // selectTransferEncoding scans content for non-ASCII characters and selects 'b' or 'q' encoding.
-func selectTransferEncoding(content []byte, quoteLineBreaks bool) transferEncoding {
+func (p *Part) selectTransferEncoding(content []byte, quoteLineBreaks bool) transferEncoding {
 	if len(content) == 0 {
```
jhillyerd, line 307:
```suggestion
	if len(content) == 0 || p.encoder != nil && p.encoder.enforceQuotedCte {
```
I think most of the changes you made to this func are not needed, we don't care about a threshold if we are always selecting te7Bit?

jhillyerd (changes requested):
I like adding the Encoder options.

I don't accept PRs without tests, as subsequent PRs to the same code often end up breaking or reverting that feature.  Take a look at this test as a possible starting point:
https://github.com/jhillyerd/enmime/blob/1d9fc9cfdfa5080fc6c0235c57166fab72c56671/encode_test.go#L327

jhillyerd:
Looks pretty good to me, not sure if you're ready for review since it's still marked as draft?

pgradoboev:
James, I wanted to add more tests after the holidays. But if you think that's enough, then I've changed the request status.

jhillyerd (approved):
Made a tweak to the test to check a bit of the encoded content.

PR #360, opened 2025-01-02
fix: Handle spaces inside angle brackets

### LINKED ISSUE #339: Possible to handle these `From:` values?
I got messages with `From:` like this:
```
From: Name< user@domain.com >
From: Name <sales@a.com >
```
Both report `unclosed angle-addr` error, but MUAs work fine.

I wonder whether we can handle this in enmime. :)

dflse:
closes #339

### DIFF
--- a/internal/stringutil/addr.go
+++ b/internal/stringutil/addr.go
@@ -34,6 +34,7 @@ func EnsureCommaDelimitedAddresses(s string) string {
 	inQuotes := false
 	inDomain := false
 	escapeSequence := false
+	inAngles := false
 	sb := strings.Builder{}
 	for i, r := range s {
 		if escapeSequence {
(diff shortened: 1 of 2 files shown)


internal/stringutil/addr.go
```diff
@@ -74,13 +88,19 @@ func EnsureCommaDelimitedAddresses(s string) string {
...
+				if r == ' ' && !inAngles {
 					inDomain = false
 					sb.WriteRune(',')
 					sb.WriteRune(r)
 					continue
 				}
 			}
+
+			if inAngles {
+				if r == ' ' {
```
jhillyerd, line 100: Could just use `&&` for these two ifs

jhillyerd (changes requested):
Looks good, one change

jhillyerd (approved):

jhillyerd:
Thanks!

iredmail:
Thanks @dflse :)

PR #372, opened 2025-04-20
Refactor EnvelopeFromPart()'s scope

bgedney:
## What
I've extracted the functionality from EnvelopeFromPart which gathered errors from any/all nested/child/sibling parts into a new method on enmime.Envelope called GatherNestedErrors().  GatherNestedErrors() is also implemented in ReadEnvelope().

## Why
The scope of the current form of EnvelopeFromPart is inconsistant in that the function's supposed purpose is to create a header from the root of a provided part, but the concluding body of the current form of EnvelopeFromPart() gathering errors from any/all nested/child/sibling parts.  This is an effort to clarify the scope of work of EnvelopeFromPart to focus on any single part.

## Testing
All current tests pass, and because the new method GatherNestedErrors() is also called ReadEnvelope(), ReadEnvelope()'s tests cover the GatherNestedErrors().

### DIFF
--- a/envelope.go
+++ b/envelope.go
@@ -159,7 +159,18 @@ func (p Parser) ReadEnvelope(r io.Reader) (*Envelope, error) {
 	if err != nil {
 		return nil, errors.WithMessage(err, "Failed to ReadParts")
 	}
-	return p.EnvelopeFromPart(root)
+
+	e, err := p.EnvelopeFromPart(root)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


jhillyerd:
This seems reasonable to me.  If you continue with this, please also update the doc comments, and add a test enforce EnvelopeFromPart not gathering errors.

jhillyerd:
Please re-open or resubmit if you pick this back up.

PR #392, opened 2026-04-07
feat!: Refactor EnvelopeFromPart()'s scope

bgedney:
### Summary
Refactor EnvelopeFromPart()'s scope so that the functions's scope of work is more aligned to the root aspect of a provided part:

- Extracted the functionality from EnvelopeFromPart which gathered errors from any/all nested/child/sibling parts into a new method on enmime.Envelope called GatherNestedErrors(). GatherNestedErrors() is also implemented in ReadEnvelope().
- Also extracted the previous functionality into a new method on the Envelope type - GatherNestedErrors(). Also implemented GatherNestedErrors() within ReadEnvelope() after the call to EnvelopeFromPart() and added supporting test cases.

### Why
The scope of the current form of EnvelopeFromPart is inconsistant in that the function's supposed purpose is to create a header from the root of a provided part, but the concluding body of the current form of EnvelopeFromPart() gathering errors from any/all nested/child/sibling parts. This is an effort to clarify the scope of work of EnvelopeFromPart to focus on any single part.

### Testing
All current tests pass, and because the new method GatherNestedErrors() is also called ReadEnvelope(), ReadEnvelope()'s tests cover the GatherNestedErrors().

#### Note:
This is me actually following through on #372 - Better late than never.

### DIFF
--- a/envelope.go
+++ b/envelope.go
@@ -175,7 +175,7 @@ func (p Parser) ReadEnvelope(r io.Reader) (*Envelope, error) {
 
 // EnvelopeFromPart uses the provided Part tree to build an Envelope, downconverting HTML to plain
 // text if needed, and sorting the attachments, inlines and other parts into their respective
-// slices.  Errors are collected from all Parts and placed into the Envelopes Errors slice.
+// slices.  Gather errors from all parts with Envelope.GatherNestedErrors().
 func EnvelopeFromPart(root *Part) (*Envelope, error) {
 	return defaultParser.EnvelopeFromPart(root)
 }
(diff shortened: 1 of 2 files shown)


envelope.go
jhillyerd, line 178: I think this comment is out of date now?  Should probably also mention they can use GatherNestedErrors if needed

jhillyerd (changes requested):
just needs a comment fix

envelope.go
bgedney: Update EnvelopeFromPart's Comment.

jhillyerd:
Thanks!

PR #398, opened 2026-06-06
fix: apply CVE-2026-42507 escaping to internal/textproto

jhillyerd:
Follow-up to #397. Now that we have local `ProtocolError` and `Error` types, apply the same CVE-2026-42507 fix that upstream applied in Go 1.26 ([CL 777060](https://go-review.googlesource.com/c/go/+/777060)):

- `Error.Error()` now quotes `Msg` via `%q` to prevent injection of misleading content or terminal control bytes
- All `ProtocolError` messages that include user-controlled input now use `fmt.Sprintf` with `%q` instead of raw string concatenation

This ensures enmime is not a vector for the vulnerability regardless of which Go version it is built with.

### DIFF
--- a/inspect_test.go
+++ b/inspect_test.go
@@ -4,12 +4,12 @@ import (
 	"bytes"
 	"io"
 	"net/mail"
-	"net/textproto"
 	"strings"
 	"testing"
 
 	"github.com/jhillyerd/enmime/v2"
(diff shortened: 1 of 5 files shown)


jhillyerd:
obsoleted by #398

PR #401, opened 2026-06-07
fix: assemble RFC 2231 multi-segment filename params with non-UTF-8 charsets

### LINKED ISSUE #109: Incorrect parsing in Filename with the following structure
**What I did: **

If the filename is too long, it is stored separately.
But when I call (*enmime.Part).FileName, it only reads the last FileName.

example)
```
--------------D5FCD174A36A19BAC239EB64
Content-Type: application/x-zip-compressed;
 name="=?UTF-8?B?6rCc7J247KCV67O0X+qwgTMw6rCcLnppcA==?="
Content-Disposition: attachment;
 filename*0*=euc-kr''%B0%B3%C0%CE%C1%A4%BA%B8%5F%B0%A2%33%30%B0%B3%2E%7A%69;
 filename*1*=%70
```
In the above example, the filename is the sum of filename*0* and filename*1*, but only filename*1* is read.
Results: p (%70)

**What I expected:**
sum of filename*0* and filename*1*

**What I got:**
only value of filename*1*

**Release or branch I am using:**
both master and release v0.4.0

troclaux:
## Summary

Go's `mime.ParseMediaType` only supports UTF-8 and US-ASCII in RFC 2231 charset-encoded continuation parameters. When a `Content-Disposition` header uses a non-ASCII charset (e.g. `EUC-KR`) split across multiple `filename*N*=` segments, stdlib discards all but the last successfully decoded segment, producing a truncated filename like `p.zip` instead of the full Korean filename.

## Changes

- Add `assembleRFC2231Params` in `mediatype/mediatype.go` to detect RFC 2231 continuation parameters (`name*N*=` / `name*N=`), sort them by segment number, concatenate the percent-decoded bytes, and convert from the declared charset to UTF-8 using enmime's existing `coding.ConvertToUTF8String`.
- The assembled value is re-encoded as a single-segment `name*=utf-8''...` parameter before being passed to `mime.ParseMediaType`, which stdlib handles correctly.
- Add unit tests in `mediatype/mediatype_test.go` for both EUC-KR and US-ASCII multi-segment cases.
- Add an integration test in `part_test.go` and test data `testdata/parts/long-filename-euckr.raw` to verify the `FileName` field is set correctly end-to-end.

## Testing

- All existing tests pass (`go test ./...`)
- New tests cover the EUC-KR multi-segment case (`개.txt` from segments `filename*0*=euc-kr''%B0%B3` + `filename*1*=%2E%74%78%74`)
- New test covers US-ASCII multi-segment continuation

Fixes #109

### DIFF
--- a/mediatype/mediatype.go
+++ b/mediatype/mediatype.go
@@ -3,7 +3,10 @@ package mediatype
 import (
 	"fmt"
 	"mime"
+	"net/url"
+	"regexp"
 	"slices"
+	"strconv"
 	"strings"
 	_utf8 "unicode/utf8"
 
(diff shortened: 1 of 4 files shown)


mediatype/mediatype.go
```diff
@@ -48,9 +51,157 @@ func Parse(ctype string) (mtype string, params map[string]string, invalidParams
...
 // ParseWithOptions parses media-type with additional options controlling the parsing behavior.
+// rfc2231SegmentRe matches RFC 2231 continuation parameter names such as "filename*0*" or "filename*1".
+var rfc2231SegmentRe = regexp.MustCompile(`(?i)^([a-zA-Z0-9!#$&\-^_.+]+)\*(\d+)(\*?)$`)
+
+// assembleRFC2231Params detects RFC 2231 multi-segment continuation parameters (paramname*N*= or
+// paramname*N=) in a media type or content-disposition string, assembles their values in segment
+// order, applies charset conversion, and returns a new string where the continuation parameters
+// are replaced by a single decoded parameter. This is needed because Go's standard
+// mime.ParseMediaType only supports UTF-8 and US-ASCII in RFC 2231 charset-encoded parameters.
+func assembleRFC2231Params(s string) string {
```
jhillyerd, line 62: `s` param name isn't very descriptive, maybe use `ctype` to align with our other funcs

mediatype/mediatype.go
```diff
@@ -48,9 +51,157 @@ func Parse(ctype string) (mtype string, params map[string]string, invalidParams
 }
 
 // ParseWithOptions parses media-type with additional options controlling the parsing behavior.
+// rfc2231SegmentRe matches RFC 2231 continuation parameter names such as "filename*0*" or "filename*1".
+var rfc2231SegmentRe = regexp.MustCompile(`(?i)^([a-zA-Z0-9!#$&\-^_.+]+)\*(\d+)(\*?)$`)
```
jhillyerd, line 55: Specifying `(?i)` and both a-z & A-Z is redundant

mediatype/mediatype_test.go
```diff
@@ -523,6 +523,18 @@ func TestParseMediaType(t *testing.T) {
 			mtype:  "application/pdf",
 			params: map[string]string{"name": "key=value"},
 		},
+		{
```
jhillyerd, line 526: the code coverage report isn't loading, but these tests look like they just cover the happy path.  We should make sure we are triggering some of the failure/continue paths and confirming reasonable output.  Doesn't need to be in this particular test func.

jhillyerd (changes requested):

mediatype/mediatype.go
```diff
@@ -48,9 +51,157 @@ func Parse(ctype string) (mtype string, params map[string]string, invalidParams
...
+				decoded, err := url.PathUnescape(val)
+				if err != nil {
+					failed = true
+					break
+				}
+				rawBytes = append(rawBytes, decoded...)
+			} else {
+				rawBytes = append(rawBytes, val...)
+			}
+		}
```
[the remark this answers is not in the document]
jhillyerd: I'm not sure how we should handle non-contiguous segments, enmime tries to be lenient nowdays... but only respecting charset on seg#0 may be a good idea?

jhillyerd:
Thanks for sending this.  Sorry for the Copilot review, but I don't spend much time reading Go code nowdays, need all the help I can get. :)

I think you may also want to run go fix, the enmime codebase has been "fixed" recently so we are using some of the newer range syntaxes etc.

PR #404, opened 2026-07-10
Ensuring returned `error`s include a stack

dcormier:
Some returned errors didn't include stacks, making it more challenging to find where things weren't going right.

Now they do.


jhillyerd (approved):

jhillyerd:
Thanks!

Issue #162, opened 2020-08-15
Remove repeated parameters in fixMangledMediaType


int01:
Hi,

Function fixMangledMediaType from header.go does not remove repeated parameters in some situations.

For example from this line

`text/html;charset=UTF-8; charset=UTF-8`

because space is not present before first charset.

You can fix this by changing:

```
			pair := strings.SplitAfter(p, "=")
			if strings.Contains(mtype, pair[0]) {
				// Ignore repeated parameters.
				continue
			}
```

to this

```
			pair := strings.SplitAfter(p, "=")
			if strings.Contains(mtype, strings.TrimSpace(pair[0])) {
				// Ignore repeated parameters.
				continue
			}
```

Thanks.

jhillyerd:
Hey man, not trying to be rude, but you are using GitHub wrong.  There is an editor built in, and sending a PR will automatically run our unit tests, so you can do all of this without even having a Go development environment.  I just created #163 using that method to make sure it works.

PR #406, opened 2026-07-20
mediatype: dedup parameters by exact name, not substring

gaoflow:
`fixMangledMediaType` deduplicates media-type parameters (added for #162) with a substring test:

```go
if strings.Contains(mtype, strings.TrimSpace(pair[0])) { // pair[0] == "name="
    continue
}
```

`mtype` is the parameter section accumulated so far, so a *distinct* parameter is dropped whenever its `"<name>="` token appears in the earlier text — usually when the name is a suffix of an earlier one (`name` after `filename`, `id` after `uuid`, `type` after `subtype`), or when `"<name>="` occurs inside an earlier value. Since `filename` and `name` co-occur in most attachment headers, `application/octet-stream; filename="a"; name="b"` loses `name` and the attachment filename is silently discarded.

These are all well-formed RFC 2045 headers that Go's `mime.ParseMediaType` returns intact, so it's a clean oracle for the intended result:

```
application/octet-stream; filename="a.txt"; name="b.txt"   -> name dropped
text/x; uuid="u"; id="7"                                    -> id dropped
text/plain; note="size=5"; size="10"                        -> size dropped
```

Fix: track seen parameter names in an exact, case-insensitive set (RFC 2045 §5.1 makes names case-insensitive) and skip only true duplicate keys. Exact repeats still collapse (the #162 behavior); distinct parameters are kept.

I diffed a broader well-formed battery (varied param order, quoted values containing `;`/`=`, RFC 2231 extended params, encoded-words) against `mime.ParseMediaType` — the substring dedup was the only divergence; `fixUnquotedSpecials`, `fixUnescapedQuotes` and `removeTrailingHTMLTags` matched stdlib on every well-formed input. (#401 edits the same file but a different seam — RFC 2231 multi-segment reassembly — not this dedup.)

Added `TestParseMediaType` / `TestFixMangledMediaType` cases plus an end-to-end `ReadParts` test asserting the attachment `FileName` survives. `go test ./...` passes.

### DIFF
--- a/mediatype/mediatype.go
+++ b/mediatype/mediatype.go
@@ -82,6 +82,7 @@ func fixMangledMediaType(mtype string, sep rune, options ParseOptions) string {
 
 	parts := stringutil.SplitUnquoted(mtype, sep, '"')
 	mtype = ""
+	seen := map[string]bool{}
 	if strings.Contains(parts[0], "=") {
 		// A parameter pair at this position indicates we are missing a content-type.
 		parts[0] = fmt.Sprintf("%s%s %s", ctAppOctetStream, strsep, parts[0])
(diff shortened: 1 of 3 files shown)


jhillyerd:
Nice, thanks

Issue #147, opened 2020-02-21
Delete child part


jhillyerd:
I'm not sure what the use cases are, but may be worth adding this feature:

https://github.com/igorrendulic/enmime/commit/e86527c50b9ccd8867f98fe87932e02e470080cc

PR #411, opened 2026-08-26
Add Part.DeleteChild to remove a child from the tree

ChrisJr404:
This adds a DeleteChild method on Part to go with the existing AddChild, per #147.

AddChild lets you build up the part tree, but there is no way to take a child back out. So if you parse a message and want to drop a part (an attachment, say) before re-encoding, you have to relink the sibling chain by hand. DeleteChild does that for you: it removes the child, joins the siblings on either side of it back together, and clears the removed child's Parent and NextSibling so it becomes a standalone part. It is safe to call on a nil part and is a no-op when the part passed in is not actually a direct child.

I named it DeleteChild to match DeleteHeader on Envelope, but happy to rename to RemoveChild or whatever reads best to you.

Added a unit test covering removal from the front, middle and end of the chain, plus the nil and not-a-child cases. make test and make lint both pass.


jhillyerd:
Looks good, thanks

